PR #37, opened 2025-07-10
feat: added about page, hide keyboard on drawer open, show user model…

NorbertKlockiewicz:
- added about page in settings
- keyboard hides correctly when drawer gets opened
- user models are shown on list when only featured are shown


app/model-hub.tsx
```diff
@@ -59,7 +59,10 @@ const ModelHubScreen = () => {
     const matchesSearch = model.modelName
       .toLowerCase()
       .includes(search.toLowerCase());
-    const matchesFilter = activeFilters.has('featured') ? model.featured : true;
+    const matchesFilter =
+      activeFilters.has('featured') && model.source === 'built-in'
```
jakmro, line 63: Maybe we can use TS enums instead of hardcoded values like 'built-in' in the future.

jakmro (approved):

PR #56, opened 2025-07-25
fix: avoid layout shifts when focusing text inputs

### LINKED ISSUE #49: Too wide text input border
When selecting text input their border width increases and shifts rest of the screen content.

wjkg:
Fixes #49 
- applies border styles to a separate view that overlaps the text field, so border width changes don't influence layout
- extracted to a component to share the styles between `TextFieldInput`, `BottomSheetTextInput`, and `TextAreaField`

| <video src="https://github.com/user-attachments/assets/b46353ee-50c8-4835-b8d7-192bd1173e0a" /> | <video src="https://github.com/user-attachments/assets/499a84bd-a62c-4aa0-972f-ada66334f471" /> |
| --- | --- |

### DIFF
--- a/components/TextAreaField.tsx
+++ b/components/TextAreaField.tsx
@@ -3,6 +3,7 @@ import { TextInput, StyleSheet, View, TextInputProps } from 'react-native';
 import { useTheme } from '../context/ThemeContext';
 import { fontFamily, fontSizes } from '../styles/fontStyles';
 import { Theme } from '../styles/colors';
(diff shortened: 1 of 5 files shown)


mkopcins:
@wjkg I know this is not the scope of this ticket but can you please do the same to the labels on model selection screen? When you have no labels selected and then select the first label it shifts layout in the same fassion

mkopcins (approved):

PR #61, opened 2025-07-29
feat: Scroll to newly downloaded model

### LINKED ISSUE #48: Model moves when selected
When selecting a model in the model selection list it moves to the top efectively disappearing from the screen. Change it to smooth scroll to the top or simply mark it as selected and only filter the list next time it is loaded

wjkg:
Fixes #48 

- keeps model that starts (or stops) being downloaded at the original position
- model that finishes downloading is moved to the top, ~and the list scrolls to show it in the new place~
  - to avoid repetition, the default "standard model list" (divided in downloaded and available sections) is now rendered as special case of grouped models.


wjkg:
PR with downloaded filter: #79

PR #69, opened 2025-08-01
fix: correct labels in local model forms

### LINKED ISSUE #67: Fix form labels for local modals
When adding or editing a local model, the fields should be labelled as "Model file" instead of "Model URL" etc.

wjkg:
Changes labels on add and edit local model screens to use "file" instead of "URL"

Fixes #67

### DIFF
--- a/app/modal/edit-local-model/[modelId].tsx
+++ b/app/modal/edit-local-model/[modelId].tsx
@@ -102,7 +102,7 @@ export default function EditLocalModelScreen() {
             />
           </View>
           <View style={styles.textFieldSection}>
-            <Text style={styles.label}>Model URL</Text>
+            <Text style={styles.label}>Model file</Text>
             <UploadInput
               fileInfo={localModelPath}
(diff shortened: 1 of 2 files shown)


app/modal/edit-local-model/[modelId].tsx
```diff
@@ -102,7 +102,7 @@ export default function EditLocalModelScreen() {
             />
           </View>
           <View style={styles.textFieldSection}>
-            <Text style={styles.label}>Model URL</Text>
+            <Text style={styles.label}>Model file</Text>
```
Comment (NorbertKlockiewicz) [inline app/modal/edit-local-model/[modelId].tsx:105]:
```suggestion
            <Text style={styles.label}>Model File</Text>
```

Let's do `file` uppercase, I think it looks better, also in other places we are doing it like that, e.g `Model Name`

NorbertKlockiewicz (changes requested):

NorbertKlockiewicz (approved):

PR #70, opened 2025-08-01
fix: do not keep screens on stack after navigating away via the drawer

### LINKED ISSUE #64: Duplicate screens after switching between drawer options
When selecting one of the screens listed in drawer menu (new chat, models, existing chats), a new screen is always pushed on top, instead of replacing the currently open one. This can result in multiple copies of e.g. model hub screen.

This has no visible effect, but can be checked by inspecting the components or navigation state:

Image

wjkg:
Previously, selecting an item from the drawer would push a new screen on the stack. This commit changes the drawer to replace the screen instead.
- ~the history of visited screens is recorded to make the Android back button behave the same as before~
    simplified to always go back to index route ([default option](https://reactnavigation.org/docs/drawer-navigator#backbehavior) for React Navigation drawer navigator)
~- navigating to drawer screens from somewhere else can cause gaps in history, but currently it happens only when creating a new chat, which is done on the initial screen.~ 

Fixes #64

### DIFF
--- a/components/drawer/DrawerMenu.tsx
+++ b/components/drawer/DrawerMenu.tsx
@@ -1,6 +1,6 @@
-import React, { useMemo } from 'react';
-import { Text, StyleSheet, View } from 'react-native';
-import { useRouter, usePathname } from 'expo-router';
+import React, { useEffect, useMemo, useRef } from 'react';
+import { Text, StyleSheet, View, BackHandler } from 'react-native';
+import { useRouter, usePathname, useNavigation } from 'expo-router';
 import { ScrollView } from 'react-native-gesture-handler';
 import { useTheme } from '../../context/ThemeContext';
 import { useChatStore } from '../../store/chatStore';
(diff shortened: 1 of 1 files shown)


components/drawer/DrawerMenu.tsx
```diff
@@ -52,19 +52,60 @@ const DrawerMenu = ({ onNavigate }: { onNavigate: () => void }) => {
...
     [...chats].sort((a, b) => b.lastUsed - a.lastUsed)
   );
 
+  const history = useRef<string[]>(['/']);
+
   const navigate = (path: string) => {
     interrupt();
-    pathname === path ? router.replace(path) : router.push(path);
+    if (history.current.at(-1) !== pathname) {
+      history.current.push(pathname);
```
wjkg, line 69: The part with the history is added only to avoid unexpectedly changing the older behaviour, but if keeping it the same is not necessary, this could be simplified by e.g. always going back to index, like it's done in drawer navigator by default.

NorbertKlockiewicz: I think it's not necessary to keep old behaviour, let's simplify it.

NorbertKlockiewicz (approved):

PR #92, opened 2025-08-07
fix: can't create a room with different model than previously selected

NorbertKlockiewicz:
This PR:
- fixes an error when model set for a room didn't update when new room was created

### DIFF
--- a/app/(drawer)/chat/[id].tsx
+++ b/app/(drawer)/chat/[id].tsx
@@ -18,18 +18,22 @@ export default function ChatScreenWrapper() {
 
   const chatId = parseInt(rawId);
   const chat = getChatById(chatId);
-  const resolvedModelId = modelId ?? chat?.modelId;
-  const chatModel = resolvedModelId
-    ? getModelById(parseInt(resolvedModelId.toString()))
-    : undefined;
 
-  const [model, setModel] = useState<Model | undefined>(chatModel);
+  const [model, setModel] = useState<Model | undefined>(undefined);
   const [isLoading, setIsLoading] = useState(true);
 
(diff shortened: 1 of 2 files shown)


app/(drawer)/chat/[id].tsx
```diff
@@ -18,19 +18,23 @@ export default function ChatScreenWrapper() {
...
   });
 
+  useEffect(() => {
+    const resolvedModelId = modelId ?? chat?.modelId;
+    if (resolvedModelId) {
+      const newModel = getModelById(parseInt(resolvedModelId.toString()));
+      setModel(newModel);
+    }
+  }, [modelId, chat?.modelId, getModelById]);
+
```
Comment (wjkg) [inline app/(drawer)/chat/[id].tsx:37]: As I understand it, the bug is caused by the router only updating params when switching between chats instead of remounting the screen component. This works, but it's not the fix I'd recommend — patching it over with additional effect makes the logic harder to follow.

Instead, have you considered resetting the state on param change? i.e. wrap it in a component that derives a key from screen params:

```typescript
export default function OuterChatScreen() {
  const { id, modelId } = useLocalSearchParams<ChatScreenParams>();

  return (
    <ChatScreenWrapper
      key={`${id},${modelId ?? ''}`}
      id={id}
      modelId={modelId}
    />
  );
}

function ChatScreenWrapper({ id: rawId, modelId }: ChatScreenParams) {
  // rawId, modelId can be expected to be constant
  ...
}
```

(that is, the issue falls exactly under [this case](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) from React docs)

wjkg (approved):

PR #98, opened 2025-08-14
feat: rag implementation

NorbertKlockiewicz:
- added react-native-rag and react-native-pdfium as dependencies
- new sources(sources.tsx) page with a list of added resources and bottom sheet(SourceManagementSheet.tsx) with possibility to rename and remove source, the list consists of new component(SourceCard.tsx)
- added new component to chatBar(ChatBarActions.tsx) which will be a container for sources, think, send/interrupt message, start voice recording buttons
- created vectorStoreContext(VectorStoreContext.tsx) to access vectorStore from different components
- new store(sourcesStore.ts) to share and manage state releated to sources between sources and source management sheet
- new repository and table in database for sources (sourcesRepository.tsx)
- BottomSheetSearchInput was extracted to separate component as it’s used both in SourceSelectSheet.tsx and ModelSelectSheet.tsx
- new bottom sheet(SourceSelectSheet.tsx) added to chat screen, it consists of few new components(ActiveSourcesSection.tsx, EmptySourcesSection.tsx, SourceListSection.tsx, ActiveSourceCard.tsx) allowing to include source in chat context
- Added PhantomChat to chatStore which represents the chat state before the first message is sent, it allows us to include sources in the chat without adding the records in database, that’s why the context for llm is prepared in ChatScreen instead of llmStore
- Added new icons
- New useSourceUpload.ts hook with a logic to upload source files
- also some styling improvements in other components


components/bottomSheets/SourceSelectSheet.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const { theme } = useTheme();
+  const { sources } = useSourceStore();
+  const { enableSource } = useChatStore();
+  const { sendEventMessage } = useLLMStore();
+  const db = useSQLiteContext();
+  const { uploadSource } = useSourceUpload();
+  const styles = useMemo(() => createStyles(theme), [theme]);
+  const [search, setSearch] = useState('');
+  const [activeSources, setActiveSources] = useState<Source[]>([]);
+  const [deactivatedSources, setDeactivatedSources] = useState<Source[]>([]);
```
wjkg, line 46: `enableSource` should already update the store where `enabledSources` are coming from. Do we need the local copy of these two lists for something else?

components/chat-screen/MessageItem.tsx
```diff
@@ -76,54 +75,73 @@ const MessageItem = memo(
...
-                </Text>
-              )}
+        {role === 'event' ? (
+          <View style={styles.eventMessage}>
+            <Text style={styles.eventMessageFileName}>
+              {content.split(' ')[0]}{' '}
+              <Text style={styles.eventMessageText}>
+                {content.slice(content.indexOf(' ') + 1)}
+              </Text>
+            </Text>
```
wjkg, line 85: small edge case:

image

ios/PrivateMind.xcodeproj/project.pbxproj
```diff
@@ -394,7 +394,7 @@
 					"-lc++",
 				);
 				OTHER_SWIFT_FLAGS = "$(inherited) -D EXPO_CONFIGURATION_DEBUG";
-				PRODUCT_BUNDLE_IDENTIFIER = com.swmansion.privatemind;
+				PRODUCT_BUNDLE_IDENTIFIER = com.swmansion.privatemind325434234434532332fe;
```
wjkg, line 397: this seems accidental?

components/chat-screen/ChatScreen.tsx
```diff
@@ -1,60 +1,88 @@
...
+
+const prepareContext = async (
+  prompt: string,
+  enabledSources: number[],
+  vectorStore: OPSQLiteVectorStore
+) => {
+  let context = await vectorStore.similaritySearch(
+    prompt,
+    K_DOCUMENTS_TO_RETRIEVE
+  );
```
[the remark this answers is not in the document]
NorbertKlockiewicz: Hmm, that's some internal executorch error, I will look into this but probably can't be fixed from here. I will add fallback strategy so the messages won't be lost.

wjkg (approved):

PR #109, opened 2025-08-22
feat: speech-to-text in chat

wjkg:
Adds a feature for creating chat messages by speaking.
- installs `react-native-audio-api` and configures recording permissions
- adds icon in chat text input to start recording
- adds visualization of microphone input and preview of transcribed text
- bundles files for `WHISPER_TINY_EN`

To be added, possibly in a separate PRs:
- ~enable for Android — the feature is currently visible only on iOS, needs #105 or the model loading issue to be resolved for 0.4~
- ~use bundled model (also needs #105 to work) or handle model download (warn about metered/unavailable connection)~
- transition between regular chat text input and speech-to-text UI - postponed due to expected conflicts with #98 
- ~fix no reaction when using speech input for the initial message in chat~


components/chat-screen/ChatSpeechInput.tsx
```diff
@@ -0,0 +1,315 @@
...
+      animationRef.current?.pushChunk(data);
+    },
+  });
+
+  const onSubmit = useStableCallback(onSubmitProp);
+  const onCancel = useStableCallback(onCancelProp);
+
+  useEffect(() => {
+    const startListening = async () => {
+      const permissionStatus = await AudioManager.requestRecordingPermissions();
```
NorbertKlockiewicz, line 66: Maybe we can request audio permissions earlier, for example on button click. Right now there's a long latency between clicking the speech to text button and alert showing.

I would suggest the flow to be button click -> request permissions -> load the model

NorbertKlockiewicz (changes requested):

NorbertKlockiewicz (approved):

PR #110, opened 2025-08-25
deps: upgrade react-native-executorch to 0.5.1

wjkg:
- installs react-native-executorch 0.5.1
- switches from `responseCallback` to `tokenCallback` when loading LLM


NorbertKlockiewicz (approved):
Tested on:
- iOS simulatr
- iOS device
- android simulator
- android device

Works on every device.

PR #115, opened 2025-08-26
fix: loading models when offline

### LINKED ISSUE #113: Remote models don't load
Attempting to write in a chat with custom remote model results in an error:

```
Error loading model: [Error: Cannot read property 'source' of undefined]
```

wjkg:
Fixes #113, which turned out to be broader — downloaded models fail to load when the app is offline. To reproduce:

1. download a model
2. disable internet connection
3. open a chat with this model
4. send a message

It's solved by loading the models via the paths of files downloaded by `ResourceFetcher` instead of the original urls. 

(likely caused by changes in fetch utils in `react-native-executorch` after upgrading to 0.5.x)

### DIFF
--- a/store/llmStore.ts
+++ b/store/llmStore.ts
@@ -14,6 +14,10 @@ import { BenchmarkResultPerformanceNumbers } from '../database/benchmarkReposito
 import { type Message as ExecutorchMessage } from 'react-native-executorch';
 import { Platform } from 'react-native';
 import { Feedback } from '../utils/Feedback';
+import {
+  ResourceFetcherUtils,
+  RNEDirectory,
+} from '../fetchUtils/ResourceFetcherUtils';
 
 interface LLMStore {
   isLoading: boolean;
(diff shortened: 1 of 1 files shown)


store/llmStore.ts
```diff
@@ -14,6 +14,10 @@ import { BenchmarkResultPerformanceNumbers } from '../database/benchmarkReposito
 import { type Message as ExecutorchMessage } from 'react-native-executorch';
 import { Platform } from 'react-native';
 import { Feedback } from '../utils/Feedback';
+import {
```
NorbertKlockiewicz, line 17: Let's completely remove our copy of ResourceFetcher and use the one provided with React Native ExecuTorch as it's the same implementation

NorbertKlockiewicz:
New version of rne was published, let's bump it in this PR and see if everything is working and if above changes are still needed

wjkg:
Not needed for loading models offline after react-native-executorch upgrade. Closing in favor of #127

PR #120, opened 2025-08-28
fix: sidebar labels cut off too early

wjkg:
Sometimes labels in the drawer menu are shortened with ellipsis even when there's enough space:

image


NorbertKlockiewicz:
I think it's some expo/react-native bug in a dev mode but I am not 100% sure

PR #128, opened 2025-08-29
feat: enable/disable thinking

NorbertKlockiewicz:
## Summary

This PR implements per-chat thinking functionality for AI models that support it, allowing users to enable/disable the thinking mode on a per-chat basis with proper validation and UI feedback.

### Key Features

• **Database Schema Updates**: Added `thinking` column to models table and `thinkingEnabled` column to chat settings with backward-compatible migrations
• **Per-Chat Thinking Memory**: Each chat remembers whether thinking is enabled or disabled for thinking-capable models  
• **UI Toggle Interface**: Added thinking toggle button in chat bar with visual indicators (light bulb icons)
• **Model Validation**: Prevents enabling thinking on models that don't support it with toast notifications
• **Empty Thinking Block Filtering**: Thinking blocks are only displayed when they contain actual content
• **Seamless Integration**: Uses existing `useChatSettings` hook for state management

### Technical Implementation

- **Migration Logic**: Automatically detects and adds missing database columns, updates default models with thinking capabilities
- **State Management**: Integrates with phantom chat system and persistent database storage
- **Error Handling**: Toast notifications for invalid toggle attempts with clear user messaging
- **Component Architecture**: Clean separation of concerns with thinking logic in ChatScreen and UI components in ChatBar

### Testing Considerations

- ✅ Database migrations work on existing databases
- ✅ New installs create proper schema
- ✅ Thinking toggle works for supporting models
- ✅ Toast shows for non-supporting models
- ✅ Empty thinking blocks are hidden
- ✅ Settings persist across app restarts
- ✅ Phantom chats maintain thinking state


database/db.ts
wjkg, line 1: Something may be wrong with the migration — Installing the app on top of the `main` version fails to load with

```
 ERROR  Warning: Error: Calling the 'prepareAsync' function has failed
→ Caused by: Error code 1: table models has no column named thinking
```

Clean install works fine.

PR #144, opened 2025-09-03
feat: Add prompt suggestions (#102)

### LINKED ISSUE #102: Add hardcoded prompts
E.g how can I cook eggs, how to calculate something, design can be similar to cards over text input in grok application, designs should be discussed with designer

Image

Amang9446:
- Add PromptSuggestions component with horizontal scrollable cards
- Move prompt suggestions from Messages to ChatBar above TextInput
- Add default prompt suggestions with 10 different categories
- Implement prompt selection that populates input field
- Add proper model loading when prompts are selected
- Showing suggestions only when no messages exist
- Fixes #102


NorbertKlockiewicz (approved):
LGTM! Thank you for this amazing contribution <3

PR #226, opened 2026-06-28
Fix ios scrollbar

### LINKED ISSUE #84: Scroll bar displayed in wrong place on devices with iOS 17.5
https://github.com/user-attachments/assets/50b80a36-b824-4a0c-b00d-de2fe518173e

kfaracik:
# Summary
Fixes #84. Resolves an iOS 17.5 layout bug where the scroll indicator initially rendered on the left side before snapping back to the right on layout recalculation. Fixed the initial layout bounds calculation to ensure the native scroll bar snaps to the right edge instantly.

https://github.com/user-attachments/assets/16b1b9b5-9e6b-4986-8f76-6c81ec65cdb7


NorbertKlockiewicz (approved):
I don't have a device with ios 17.5 to test it but the code looks correct.

PR #240, opened 2026-07-09
Feat: customizable system prompts (personal preferences)

### LINKED ISSUE #230: [Feature Request] - Customizable system prompts
## Summary
Allow users to define a custom system prompt that applies across the entire app.

## Problem / Motivation
Currently the system prompt is fixed, so users can't tailor the model's default behavior, tone, or constraints to their use case. Setting it per-conversation is repetitive and error-prone.

## Proposed Solution
Add a setting where users can write and save a custom system prompt that is automatically applied to all **new** sessions/conversations in the app.

## Requirements
- Text input to define the system prompt in settings
- Persist the prompt across app restarts
- Apply it automatically to every new conversation
- Ability to edit, reset to default, or clear it

## Acceptance Criteria
- A user can set a custom system prompt in settings
- The prompt persists and applies app-wide
- The user can update or remove it at any time

kfaracik:
### PR Objective
This Pull Request adds customizable system prompts to the chat: users write global personal preferences (e.g., "be concise", "answer in Polish", "avoid jargon") that silently shape every conversation — the model follows them without ever quoting, acknowledging, or parroting them back.

Resolves **#230**.

---

#### Custom System Prompt
* **Persisted setting** (`store/settingsStore.ts`): A global `customSystemPrompt` stored via Zustand + persist on AsyncStorage.
* **Guard & injection** (`utils/promptUtils.ts`, `constants/prompts.ts`): `CUSTOM_PROMPT_GUARD` is prepended to the user's preference and appended to the system prompt during message assembly (`store/llmStore.ts`). Empty/whitespace-only preferences are skipped entirely — neither prompt nor guard is added.
* **Soft token budget** (`constants/settings.ts`): Character cap derived from a ~256-token × ~4 chars/token approximation, so the preference can't eat the on-device context — validated on the trimmed value.

#### Settings, Reworked
* **Settings is now a drawer screen** (`app/(drawer)/settings.tsx`, `app/(drawer)/_layout.tsx`) — a proper destination like Models, navigated via `replace`, instead of a full-screen modal.
* **Sub-pages as card-push screens** with `ios_from_right` (`app/_layout.tsx`): The Personal preferences editor (`app/(modals)/custom-system-prompt.tsx`) and App info.
* **App info moved out of the drawer** into Settings (`components/drawer/DrawerMenu.tsx`), alongside Personal preferences via a shared `SettingsRow` (`components/settings/SettingsRow.tsx`).
* **Drawer active-item logic fixed**: Drawer destinations use `replace` + path-equality highlighting; modal actions no longer show a stale active state.

#### Input Error State
* **Animated border** (`components/TextInputBorder.tsx`): Crossfades between inactive / active / error states via opacity (native driver), with error taking precedence over active.
* **TextAreaField gains error state** (`components/TextAreaField.tsx`): Added `error?: boolean` + `errorMessage?: string` props. Renders a red border and an inline message only when set — no layout change otherwise. The preferences editor uses it to flag the over-limit case.

#### Chore
* **Deprecation cleanup**: Replaced `StyleSheet.absoluteFillObject` (RN 0.83 `@deprecated`) with `StyleSheet.absoluteFill` across 4 components — same frozen object, no behavior change.


kfaracik:
### 🎥 Demos

**1. Adding Custom System Prompt**
Smooth navigation and setting up global preferences.
> [[DROP VIDEO 1 HERE]](https://github.com/user-attachments/assets/7ba52453-d9ba-47c7-838a-4bad1abdb627)

**2. Validation Error State**
UI feedback (red border & message) when exceeding the character limit.
> [[DROP VIDEO 2 HERE]](https://github.com/user-attachments/assets/d287becb-b576-43e2-901b-fe624a60f3ad)

**3. App Info Relocation**
Cleaner drawer hierarchy with `App info` moved inside `Settings`.
> [[DROP VIDEO 3 HERE]](https://github.com/user-attachments/assets/eca05022-bf30-41fa-8917-769a50d657ef)

components/TextInputBorder.tsx
```diff
@@ -1,21 +1,53 @@
-import React, { useMemo } from 'react';
-import { StyleSheet, View } from 'react-native';
+import React, { useEffect, useMemo, useRef } from 'react';
+import { Animated, StyleSheet, View } from 'react-native';
```
NorbertKlockiewicz: We are using `Animated` here but in the codebase we generally use reanimated, let's change it.

store/llmStore.ts
```diff
@@ -391,7 +392,8 @@ export const useLLMStore = create<LLMStore>((set, get) => ({
         get().activeChatMessages,
         context,
         settings,
-        currentModel
+        currentModel,
+        useSettingsStore.getState().customSystemPrompt
```
NorbertKlockiewicz, line 417: The first message after a cold start silently drops the preference.
This read happens before the store rehydrates from AsyncStorage, so customSystemPrompt is still the initial ''. 

Repro: save a preference like "always end replies with BANANA", force-quit, reopen, send immediately — the first reply ignores the preference; the second one honors it.

The hasHydrated flag added in this PR guards the editor screen but not this read — the one place it actually matters. Simplest fix: move the read below await waitForModelLoad(get) (model load takes far longer than rehydration), or gate on hasHydrated.

app/(modals)/custom-system-prompt.tsx
```diff
@@ -0,0 +1,138 @@
+import React, { useEffect, useMemo, useState } from 'react';
```
NorbertKlockiewicz, line 1: The 1024 characters gate is UI thing only on this screen, when in the future we will possibly add possibility to modify the preference from other place we can forgot about the limit. Let's also check it in the setter so it's safe to use in the future.

NorbertKlockiewicz (changes requested):

components/TextInputBorder.tsx
kfaracik: Thx, done!

app/(modals)/custom-system-prompt.tsx
kfaracik: Thanks, good point. Fixed!

store/llmStore.ts
kfaracik: Done

NorbertKlockiewicz (approved):

PR #245, opened 2026-07-16
fix(chat): prevent foreign key race on first PDF send

kfaracik:
## Problem
Sending the first message in a brand-new chat that has a PDF attachment could fail with `FOREIGN KEY constraint failed`. The message and its source links were persisted against a chat id that didn't exist yet, and the attachment's source rows could be cleaned up while the send referencing them was still in flight.

## Root cause
- `ChatScreen` created the new chat but kept persisting the message, enabling sources and calling `updateLastUsed` against the original `chatId!` placeholder instead of the id actually returned by `addChat`.
- `ChatBar.handleSend` called `clearAll()` synchronously, which ran `cleanupOrphanedSources` and deleted the source rows the in-flight send needed.

## Fix
- Thread the real persisted chat id through the whole send path: `addChat` → `setActiveChatId` → `router.replace` → `enableSource` / `updateLastUsed` / `sendChatMessage`.
- Let `onSend` be async and await it; clear the input immediately but defer source cleanup via `clearAll({ cleanupSources: false })` so sources survive until the send completes.
- Add a `cleanupSources` option (default `true`) to `useAttachment.clearAll`.

## How to test
- New chat → attach a PDF → send the first message: no FK error, the message and cited sources persist, and the chat opens on its real id.
- Regression: text-only send and voice-transcript send still clear the input and clean up orphaned sources as before.

### DIFF
--- a/components/chat-screen/ChatBar.tsx
+++ b/components/chat-screen/ChatBar.tsx
@@ -39,7 +39,7 @@ interface Props {
     userInput: string,
     imagePath?: string,
     attachments?: Attachment[]
-  ) => void;
+  ) => void | Promise<void>;
   onSelectModel: () => void;
   onSelectPrompt: (prompt: string) => void;
   ref: Ref<{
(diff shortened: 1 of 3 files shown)


NorbertKlockiewicz (approved):
lgtm

PR #246, opened 2026-07-16
feat(rag): hybrid retrieval + multilingual refinements on top of vector RAG

kfaracik:
> **Stacked on `feat/221-rag-sources`.** Base this PR against that branch, not `main`, and merge it after the base lands. The diff here is only the retrieval-quality layer (~16 files); everything else belongs to the base PR.

Turns the vector-only baseline into hybrid retrieval and adds the language refinements. Same on-device guarantee — nothing new leaves the device; this is purely a quality upgrade to how chunks are found and ranked.

## What it adds on top of the baseline

- **Keyword index** (`database/keywordIndex.ts`) — a parallel FTS5/BM25 table keyed by the same `documentId:chunkIndex` chunk id, built alongside the vector index and torn down with it. On first launch after upgrading from the vector baseline it **backfills from the existing `vectors` table**, so documents imported before this PR become keyword-searchable without a re-import.
- **Hybrid retrieval** (`utils/hybridRetrieval.ts`, replacing the vanilla `retrieve`) — semantic vector search + keyword BM25, fused with Reciprocal Rank Fusion, then re-ranked (term-coverage boost → MMR diversification → adaptive-k) before neighbor expansion.
- **Multilingual no-answer detection** — Polish refusal patterns added to the existing English ones, so a PL "brak informacji" reply cites nothing instead of over-citing.
- **`HybridRetriever`** — the app↔library boundary: a thin wrapper binding store + embeddings into one `retrieve(query, options)` call.

## Why hybrid (the evidence)

BM25 recovers exact-match recall — names, codes, rare tokens — that embeddings miss; vectors recover paraphrase. An offline eval harness (real LFM 2.5 + FTS5, frozen embeddings, ~1000 queries) confirms hybrid beats pure vector on exact and mixed queries, at parity on pure-paraphrase. RRF fuses the two rankings without tuning a per-query linear weight. The semantic floor those numbers were measured against (`0.40` + `0.25` top-keep floor) is the calibrated gate that ships in the base PR, so the measured recall is what this branch actually delivers.

> Re-run the on-device eval before merging: the keyword stemming (`utils/queryTerms.ts`) and the RRF/coverage/adaptive-k knobs are exactly what the harness measures, so any tuning here should be revalidated against it rather than trusted on unit tests alone.

## Key decisions / trade-offs

- **No cross-encoder re-ranker.** Re-ranking is RRF + term-coverage + MMR + adaptive-k — all cheap and on-device. A cross-encoder would improve precision but costs a second model and per-candidate inference on a phone.
- **`HybridRetriever` has no interface and no generic.** One implementation, one caller. Notably *not* `implements VectorStore`: the hybrid is read-only and its `ContextChunk` output drops id/embedding, so coercing it to a `QueryResult` would change results. Extract an interface when a second retriever appears.
- **Retrieval constants live in `constants/retrieval.ts`** with per-value rationale — the fusion weights, MMR lambda, coverage alpha, adaptive-k ratios and semantic floor each document where the number comes from.

## Known limitations

- **FTS5 degradation is soft.** If the native build lacks FTS5, `ensureKeywordIndex` logs a warning and hybrid search silently falls back to vector-only. Open question: leave as-is / surface a signal / fail fast — depends on whether a supported build without FTS5 exists.
- **Query processing is PL/EN-tuned.** `utils/queryTerms.ts` hardcodes Polish morphology (ł/Ł folding, stem-prefix) and the no-answer detection is PL/EN. Other languages work but retrieve measurably worse. Follow-up: a `LanguageAdapter` with a neutral fallback.

## How to test locally

1. **Exact-match win** — attach a doc containing a rare token (an error code, a surname) and query it verbatim; the chunk should surface even when semantic similarity alone would gate it out.
2. **PL refusal** — ask (in Polish) something the docs don't cover; expect a no-answer reply with zero citations.
3. **Diversity** — a spanning multi-doc query should not return five near-duplicate chunks from one file (MMR + per-file cap).
4. **FTS5 fallback** — on a build without FTS5, retrieval still returns results (vector-only) with a logged warning.
5. **Backfill after upgrade** — open a build that imported documents on the vector baseline, then upgrade to this branch; the pre-existing documents are keyword-searchable (verbatim rare-token query surfaces them) with no re-import.


NorbertKlockiewicz:
Models don't answer to first message, they answer only to 2nd one.

kfaracik:
> Models don't answer to first message, they answer only to 2nd one.

Done

https://github.com/user-attachments/assets/acbd0c2d-7658-4221-906e-dca346d6b57f

store/sourceStore.ts
```diff
@@ -102,7 +181,7 @@ export const useSourceStore = create<SourceStore>((set, get) => ({
             : s
         ),
       }));
-      return { success: true, sourceId };
+      return { success: true, sourceId, truncated };
     } catch (e) {
```
NorbertKlockiewicz, line 219: The abort path at line 148 calls `rollbackPartialSource()`, but this catch only removes the temp UI entry, it never rolls back the DB. If `embedDocument`/`vectorStore.add` throws partway through the loop the sources row from insertSource plus chunks 0..k in both the vectors table and chunk_fts index all survive. The attachment disappears from the UI, but a phantom half-imported source persists. We should rollback here too.

components/chat-screen/MessageItem.tsx
```diff
@@ -89,6 +105,43 @@ const MessageItem = memo(
...
+      const seen = new Set<string>();
+      return sourceDocuments.filter((source) => {
+        const key = sourceKey(source.documentId, source.name);
+        if (seen.has(key)) return false;
+        seen.add(key);
+        return true;
+      });
+    }, [sourceDocuments]);
+
+    const handleLinkPress = useCallback(({ url }: { url: string }) => {
```
NorbertKlockiewicz, line 121: Linking.openURL(url) fires on any URL the markdown parser extracts, with no scheme allow-list.
The link text is model-generated and, in a RAG turn, directly influenced by document content that can be attacker-controlled. A poisoned source can induce the model to emit [tap here](file:///…), a custom app deep-link, or tel:/sms: — and a single tap opens it with no confirmation.

Restrict to web/mail schemes before opening:



```suggestion
    const handleLinkPress = useCallback(({ url }: { url: string }) => {
        if (!/^https?:\/\//i.test(url) && !/^mailto:/i.test(url)) return; // not sure about this
        Linking.openURL(url).catch(() => {});
     }, []);
```

NorbertKlockiewicz (approved):
Some nits but I think it's worth fixing, besides that IMPRESSIVE WORK!

store/sourceStore.ts
kfaracik: Done

PR #250, opened 2026-07-17
fix(chat): coalesce streaming token updates via requestAnimationFrame

kfaracik:
The executorch token callback can fire many times in a single synchronous tick (notably while replaying the prompt during prefill). Appending to the store on every token scheduled one React update per token, and a burst in a single tick tripped "Maximum update depth exceeded", aborting the response.

Buffer tokens at module scope and flush the visible append at most once per animation frame, collapsing a burst into a single render. First-token bookkeeping (haptics, prefill-interrupt, firstTokenTime) stays synchronous via an in-flight counter; the buffer is reset on generation start / model load and cleared on complete/failed so a late flush can't duplicate the authoritative final text.

### DIFF
--- a/__tests__/llmStore.test.ts
+++ b/__tests__/llmStore.test.ts
@@ -88,7 +88,8 @@ beforeEach(() => {
   );
 });
 
-afterEach(() => {
+afterEach(async () => {
+  await flushFrame();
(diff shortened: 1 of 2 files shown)


NorbertKlockiewicz (approved):
lgtm

PR #251, opened 2026-07-17
Fix/scroll down button overlaps multiline input

kfaracik:
Fixes a set of chat input / scroll-to-bottom button issues and adds a smooth grow/shrink animation. Four self-contained commits:

- **Button positioning & keyboard sync** — the scroll-to-bottom button no longer slides under the input once it grows to 2–3 lines, or under the keyboard when it opens. The ChatBar is now driven by the same Reanimated keyboard values as the button (replacing `KeyboardStickyView`), so the bar and button animate frame-locked on one runtime instead of drifting apart.
- **No auto-scroll on resize** — adding or removing a line no longer forces the conversation to jump to the bottom; it only follows the latest message when the user is already at the bottom.
- **Android input height stability** — disables `includeFontPadding` on Android so the placeholder and the first typed character measure the same, removing a 1px jump on the first keystroke (iOS was already stable via `lineHeight`).
- **Smooth line grow/shrink** — the input now animates smoothly when it grows/shrinks between 1–3 lines (`LinearTransition` on the bar + a matching `withTiming` on the scroll content padding), keeping the bar, scroll content, and button in sync.

### DIFF
--- a/__mocks__/react-native-reanimated.ts
+++ b/__mocks__/react-native-reanimated.ts
@@ -32,6 +32,19 @@ const Animated = {
   createAnimatedComponent,
 };
 
+type AnimationBuilder = Record<
+  string,
+  (...args: unknown[]) => AnimationBuilder
+>;
+
+const makeAnimationBuilder = (): AnimationBuilder => {
+  const builder = new Proxy(
+    {},
+    { get: () => () => builder }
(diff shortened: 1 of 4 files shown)


NorbertKlockiewicz (approved):
lgtm

PR #254, opened 2026-07-22
feat(chat): fade messages into the background at the list edges

### LINKED ISSUE #253: Fade messages out at the top and bottom of the list
Messages are cut off mid-line where the list meets the header and where it meets the chat bar. The list also stops above the bar instead of scrolling under it, so the bar sits on a hard edge.

Add a background-coloured gradient at both edges so text dissolves instead of being clipped, and run the list full-height so messages pass underneath the chat bar.

On Android the bottom fade must reach full opacity by the top of the system navigation bar — in both gesture and three-button modes, which have different inset heights.

kfaracik:
Resolves #253.

Messages were clipped mid-line at the top of the list and at the chat bar. The list now runs the full height of the screen so messages scroll underneath the bar, and background-coloured gradients at both edges let them dissolve instead of cutting off.

- ChatBar is an absolute overlay rather than a spacer taking layout space; Messages takes its height as chatBarInset to keep the last message resting above it and to size the bottom fade.
- ChatBar reports its baseline height instead of the live one, and re-measures when the safe-area inset changes, so switching Android navigation modes or rotating does not leave a stale baseline.
- On Android the bottom ramp reaches full opacity at the top of the system navigation bar, driven entirely by insets.bottom, so gesture and three-button modes both follow without a special case.
- Fades leave a right-edge gutter so they do not paint over the scroll indicator, and are drawn below the scroll-to-bottom button so it keeps receiving taps.

The layout and the Android navigation-bar handling were confirmed on a physical device in both navigation modes. The later changes — scroll indicator gutter, style memoisation, comment cleanup — are lint, type and unit tested only, not yet re-checked on device.

### DIFF
--- a/__tests__/useKeyboardLift.test.ts
+++ b/__tests__/useKeyboardLift.test.ts
@@ -1,0 +1,71 @@
+import { renderHook } from '@testing-library/react-native';
+import { useReanimatedKeyboardAnimation } from 'react-native-keyboard-controller';
+import { useKeyboardLift } from '../components/chat-screen/useKeyboardLift';
+
+let mockInsetsBottom = 0;
+
+jest.mock('../context/ThemeContext', () => ({
+  useTheme: () => ({
+    theme: {
+      insets: { top: 0, bottom: mockInsetsBottom, left: 0, right: 0 },
+    },
+  }),
+}));
+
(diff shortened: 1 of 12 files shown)


NorbertKlockiewicz (approved):
Checked on iOS and looks 🔥

PR #257, opened 2026-07-23
chore(build): strip console.* from production bundles

### LINKED ISSUE #256: [Build] Strip console logging from production bundles
The app currently ships ~37 direct `console.warn` / `console.error` calls (plus occasional `console.log`) into release bundles. There is no crash reporting that consumes them, so in production they only add bundle weight and leak internal error messages into system logs (`adb logcat` / os_log).

Logging should be a dev-only affair:

- keep all `console.*` output in development and tests,
- strip every `console.*` call (including `console.error`) from production bundles,
- no changes at the call sites — handle it at build time (`babel-plugin-transform-remove-console` in the `env.production` block, the Expo-recommended approach).

kfaracik:
Resolves: #256

## PR Objective

Make logging a dev-only affair: every `console.*` call (including `console.error`) is stripped from production bundles at build time, while development and test behavior stays exactly as it is today. The app has no crash reporting that would consume production logs, so in release builds they only added bundle weight and leaked internal error messages into system logs (`adb logcat` / os_log).

## What changed?

* **`babel-plugin-transform-remove-console`** added as a devDependency.
* **`babel.config.js`:** new `env.production` block enabling the plugin. Metro sets `BABEL_ENV=production` for release bundles, so the transform applies only there — dev bundles and Jest (`test` env) are untouched.
* **No `exclude` list on purpose:** `console.error` is stripped too. If we ever add crash reporting, the follow-up should route errors through a logger instead of re-keeping raw `console.error`.
* **Zero call-site changes:** all ~37 existing `console.warn` / `console.error` calls across stores, hooks, and components stay as they are — the transform removes them from the release output.

## Verification

* Ran a sample module through the project's real Babel config in both env modes:
  * `development` → all of `log` / `warn` / `error` / `info` / `debug` kept,
  * `production` → all stripped.
* Full Jest suite and ESLint pass.

> **Note for local release builds:** reset the Metro cache once (`yarn start --reset-cache`) so the new transform takes effect.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -18,5 +18,12 @@ module.exports = function (api) {
         },
       ],
     ],
+    env: {
+      // Strip every console.* call (error included — there is no crash
+      // reporting to feed) from release bundles; dev and test keep them.
+      production: {
+        plugins: ['transform-remove-console'],
+      },
+    },
   };
 };
--- a/package.json
+++ b/package.json
@@ -118,6 +118,7 @@
     "@testing-library/react-native": "^13.3.3",
     "@types/jest": "29.5.14",
     "@types/react": "~19.2.10",
+    "babel-plugin-transform-remove-console": "^6.9.4",
     "eslint": "^9",
     "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-prettier": "^5.5.5",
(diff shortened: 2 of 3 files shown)


NorbertKlockiewicz (approved):
lgtm

PR #258, opened 2026-07-24
fix(chat): apply a default repetitionPenalty so no model can loop (#255)

### LINKED ISSUE #255: Generation can collapse into an endless repeated phrase — no repetition penalty configured for Qwen 3, no loop guard
## Summary

An assistant reply degenerated into an infinite repetition loop: the word "czapki, "repeated ~100 times until the generation budget ran out. The answer was unusable and there is currently nothing in the app that detects or stops this.

## Reproduction

- Model: Qwen 3 - 1.7B (quantized, 8da4w)
- Language: Polish
- Web search: ON — the grounding context came from a Nike promotions page, i.e. a page
  whose main content is a list of product categories
- Prompt: "Sprawdź aktualne promocje w oficjalnym sklepie nike"

Observed (truncated):

> 50% na wybrane produkty (zawierające: płyty, utek, ćwiczenia, odzieź, koszule, skarpetki, czapki, czapki, czapki, czapki, czapki, … )

Not yet established whether this reproduces deterministically with the same prompt + context, or how often it occurs. Frequency is unknown.

## Expected

Either the model does not enter the loop, or the app detects it and stops the generation instead of streaming hundreds of repeated tokens to the user.

## Analysis

This is neural text degeneration, not an app-logic bug. The relevant property is that the loop is **self-reinforcing** — the probability of a repeated phrase increases with each repetition (Holtzman et al., ICLR 2020; Xu et al., NeurIPS 2022). It therefore never recovers on its own, which means cutting the generation short costs nothing.

Aggravating factors specific to this case:
- 1.7B parameters, 4-bit quantized weights — quantization flattens the tail of the distribution and makes degeneration more likely;
- Polish output — the reply also contains grammar errors ("bezpłatny wysyłka", "odzieź"), i.e. the model is already off-distribution;
- the grounding context is list-shaped (a promotions page), which primes list continuation.

**Root-cause lead:** `react-native-executorch` exposes `repetitionPenalty` via `LLMModule.configure({ generationConfig })`, and we never set it. `constants/default-models.ts` forwards the generation config straight from the RNE model registry, and the registry entry for `qwen3-1.7b-quantized` contains only `{ temperature: 0.6, topP: 0.95 }`. By contrast the LFM 2.5 VL entries in the same registry explicitly ship `repetitionPenalty: 1.05` and `minP: 0.15` — so the library treats this as a per-model knob that is simply unset for Qwen 3. The native default value still needs to be confirmed (the docs state "1 disables the penalty").

## Proposed work, in order

1. **Set `repetitionPenalty` (~1.05–1.1) for the Qwen 3 entries**, overriding the registry defaults. Cheapest change, addresses the cause. Needs an A/B on real prompts to check it does not flatten legitimate repetition or hurt answer quality.
2. **Loop guard that interrupts generation.** The token callback in `store/llmStore.ts` already batches tokens, so the detector fits there and `LLMModule.interrupt()` is already wired. Trigger on *contiguous, identical* repetition only — e.g. the same ≤40-char unit ≥8 times in a row, or an identical line ≥6 times. On trigger: stop, keep what was generated, and mark the message as truncated.
3. **Near-repeat detection** via distinct-n or a compression ratio over a sliding windo of the tail, for loops that vary slightly ("25% na…", "30% na…", …) and so escape an exact-match check.
4. **Length ceiling** relative to the context-window budget as a last-resort backstop.

Explicitly out of scope: automatic silent regeneration. On-device that costs a full second generation, and if the context shape is what triggered the loop, a retry with the same context is likely to loop again. A user-facing "Try again" is the right shape if we want one.

## Risks

Legitimate repetition exists — markdown tables, numbered lists, code blocks, refrains. Thresholds must be generous and the match must require contiguous identical units; falsely truncating a correct table is worse than letting one loop through.

## Notes

Logprob-based detection is not available: `react-native-executorch`'s `generate()` exposes no per-token logprobs (already established while evaluating FLARE-style confidence signals for web search).

## Testing

- [ ] **Exercise the post-hoc loop guard on a device** (`utils/loopDetection.ts`, extracted as #311). It is the replacement for the `repetitionPenalty` that #289 reverted, and with the penalty back at `1` it is currently the only thing standing between a degenerate generation and the user — yet it has never been run in a device round. Two checks, in this order:
  - [ ] it cuts the reproduction above (Qwen 3 - 1.7B, Nike promotions page, Polish) at the first repeat instead of streaming the whole `czapki, czapki, czapki…` tail;
  - [ ] it leaves legitimate repetition alone — a markdown table, a numbered list and a recipe's ingredient list must come through untouched, per the **Risks** section above. Falsely truncating a correct table is the worse failure.

kfaracik:
## Problem

Small quantized models — notably **Qwen 3 1.7B** — can collapse into a self-reinforcing repetition loop. The reported case (#255) repeated `czapki, czapki, czapki…` ~100× and never terminated.

**Root cause:** the `react-native-executorch` registry only ships a `generationConfig` (`temperature`/`topP`) for *some* models and never sets `repetitionPenalty`. So `getGenerationConfigForModel` handed the runtime a config with **nothing to discourage repetition** for Qwen 3 and most other models.

## Fix

Apply a conservative default `repetitionPenalty` (`1.1`) to **every** model, via spread so any explicit registry value (e.g. LFM's tuned `1.05`) still wins:

```ts
export const DEFAULT_REPETITION_PENALTY = 1.1;

export const getGenerationConfigForModel = (modelPath: string) => ({
  repetitionPenalty: DEFAULT_REPETITION_PENALTY,
  ...GENERATION_CONFIG_BY_MODEL_PATH[modelPath],
});
```

One-line behavioral change — no new subsystems, migrations, or UI.

## Testing

**Unit** — `getGenerationConfigForModel` returns the default penalty for the Qwen 3 registry entry *and* for a model absent from the registry map. `46 passed`.

**On-device (iOS + Android simulators)** — with the penalty applied, a forced *"repeat czapki 200×"* prompt no longer degenerates:

| Platform | Result |
|---|---|
| iOS (iPhone 17) | penalty stops the runaway — output varies and terminates |
| Android (Pixel 10) | `czapki, czapki, czapki,... (podzielić na 200 części)` — 3× then self-abbreviates |

Also confirmed `LLMModule.configure({ repetitionPenalty })` works on the native side of both platforms without crashing.

## Notes

A larger defense-in-depth variant (runtime loop detection + `stopReason` persistence + an inline "stopped repeating" UI note) was prototyped but **deliberately not included** — on-device testing showed the penalty alone prevents the degeneration, so the extra surface area wasn't justified. It's preserved on the `fix/255-generation-loop-guard` branch as a fallback if a loop is ever observed despite the penalty.

Closes #255


NorbertKlockiewicz (approved):
lgtm

PR #260, opened 2026-07-27
refactor: introduce useThemedStyles hook, drop repeated styles boilerplate

kfaracik:
## What

Replaces the boilerplate repeated across the app for building theme-derived styles:

```tsx
const { theme } = useTheme();
const styles = useMemo(() => createStyles(theme), [theme]);
```

with a single call to a new hook:

```tsx
const { styles } = useThemedStyles(createStyles);        // styles only
const { styles, theme } = useThemedStyles(createStyles); // when theme is also read directly
const { styles } = useThemedStyles(createStyles, disabled); // parametric factory
```

- New hook: `hooks/useThemedStyles.ts` — memoizes the `StyleSheet` from a `createStyles(theme, ...args)` factory and returns `{ styles, theme }`. Variadic args cover the parametric factories (`disabled` / `selected`).
- 76 components migrated. `createStyles` definitions are left untouched.
- Adds `CLAUDE.md` documenting the convention so future work doesn't reintroduce the boilerplate.

## Why

The two-line `useTheme()` + `useMemo(createStyles)` pattern appeared 76 times. Collapsing it removes ~110 net lines, drops per-file `useMemo`/`useTheme` imports, and gives a single, consistent, memoized entry point. Files that also read `theme` directly no longer need a separate `useTheme()` call.

## Behavior

No behavior change — the returned `styles` and `theme` objects are identical to before; only how they're obtained changes.

## Out of scope (left intentionally)

- `Chip.tsx`, `TextFieldInput.tsx` — `createStyles` takes several non-theme args.
- `_layout.tsx`, `app-info.tsx`, `SpinningCircle.tsx`, `MarkdownComponent.tsx` — use `theme` without a `createStyles` StyleSheet.


NorbertKlockiewicz (approved):
lgtm

PR #264, opened 2026-07-29
fix(models): repair Bielik download

### LINKED ISSUE #262: Bielik model fails to download
The **Bielik** model can't be downloaded — the download does not complete.

Investigate why (model entry / source URL / tokenizer assets in the model list).

kfaracik:
The Bielik entry was the only default model with hand-written download URLs, pinned to resolve/main of the HF repo. The repo moved its files (bielik-v3.0-1.5B/quantized/... -> xnnpack/...), so the stored model URL started returning 404.

- source Bielik paths from the BIELIK_V3_0_1_5B_QUANTIZED constant of react-native-executorch, which is pinned to a version tag
- refresh stale built-in download paths on migration for undownloaded rows (seeding uses INSERT OR IGNORE, so existing installs kept the dead URL forever)
- re-sync built-in paths when a downloaded model's files are removed, so an immediate re-download uses current URLs
- correct Bielik modelSize to the actual file size (0.86 GiB)

Closes #262

### DIFF
--- a/__mocks__/react-native-executorch.ts
+++ b/__mocks__/react-native-executorch.ts
@@ -62,6 +62,9 @@ export const LFM2_5_VL_1_6B_QUANTIZED = makeModelConstants(
 export const LFM2_5_VL_450M_QUANTIZED = makeModelConstants(
   'lfm2.5-vl-450m-quantized'
 );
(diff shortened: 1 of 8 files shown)


NorbertKlockiewicz (approved):
works

PR #272, opened 2026-07-31
fix(chat): copy what the bubble shows (#269) + keep a sent message pinned

### LINKED ISSUE #269: Copying a message with <think> tags truncates the pasted user bubble; copied text includes hidden markup
## Problem

When a user copies an assistant reply that contains `<think>…</think>` tags (Copy button under the message) and pastes it back as a new user message, the user bubble renders almost empty — a very large part of the message is visually cut off. The copied text also silently includes markup that was hidden on screen.

Affects `main` (not the feature branches): the Copy button landed in #227 (`ec8fe09`), and the `<think>` parsing in message bubbles predates it.

## Root cause — two cooperating defects

1. **Copy copies the raw content.** `components/chat-screen/Messages.tsx` (`handleCopyMessage`) does `Clipboard.setStringAsync(message.content)` — including `<think>…</think>` blocks and `[n]` citation markers that are hidden/stripped in the rendered bubble. Copied text ≠ displayed text.
2. **User bubbles render only the text before `<think>`.** `parseThinkingContent` in `components/chat-screen/MessageItem.tsx` is applied to **every** role, and the user branch renders only `normalContent` (text before `<think>`), dropping both the think block and **everything after `</think>`** (`normalAfterThink` is only rendered for assistant messages). Since model replies usually start with `<think>`, the pasted user bubble appears nearly empty.

DB content is intact — the loss is purely visual — but the model receives the full pasted text with tags in the prompt.

## Proposed fix

- When copying an assistant message: strip `<think>…</think>` blocks and citation markers (reuse `stripCitations` and the `THINK_OPEN`/`THINK_CLOSE` constants from `constants/citations.ts` — the think parser is currently duplicated in `MessageItem.tsx` and `utils/messageSources.ts`).
- Render user messages verbatim (skip `parseThinkingContent`) — a user message never has a genuine think block.

## Other cases to cover

- Unclosed `<think>` (interrupted generation) — currently truncates everything to the end of the message.
- `[1]`/`[2]` citation markers present in copied text although stripped on screen.
- Plain-text "Source N" phrases (related: #268).
- `/think` / `/no_think` suffixes appended in `utils/promptUtils.ts` are **not** persisted — verified, no change needed there.

kfaracik:
## Why

Two independent chat bugs.

**1. Copying a reply with `<think>` tags (#269).** The Copy button put `message.content` on the clipboard verbatim — including the `<think>…</think>` block and the `[n]` citation markers that are hidden on screen. Pasting that back rendered a nearly empty user bubble: `parseThinkingContent` ran for every role, and the user branch rendered only the text *before* `<think>`, dropping the block and everything after `</think>`. The DB row was always intact — the loss was purely visual, but the model did receive the raw markup.

**2. A sent message sometimes wasn't pinned.** From the second message on, the new message occasionally stayed where it was appended instead of moving below the header, forcing a manual scroll. The pin was a single `scrollToEnd` fired from `onContentSizeChange`, but the list's maximum offset keeps moving for a few hundred milliseconds after that:

- `blankSpace` is applied as the scroll view's bottom **contentInset** (`ScrollViewWithBottomPadding` in keyboard-controller), and Reanimated commits it on the UI thread a frame or more after JS sets it;
- the inset is recomputed again once the new user row and the assistant placeholder report their heights;
- the keyboard dismissed by the send itself animates out for ~250 ms, and keyboard-controller drives its own per-frame `scrollTo` the whole time.

Whichever lands last wins, so a lone scroll aims at a stale end and falls short. The first message in a chat was masked by the initial-scroll retry loop; later ones had no safety net. Independently, the Android keyboard-dismiss re-snap was gated on `wasAtBottomDuringKeyboard`, so sending after scrolling up skipped the last correction.

## What changed

- `utils/thinking.ts` — one home for the think parser, previously duplicated in `MessageItem` and `messageSources`; also normalises an orphan `</think>`.
- `utils/messageText.ts` — `visibleMessageText()` decides what Copy puts on the clipboard: think blocks out, citation markers out when the reply is grounded in sources, and a reply that is *only* reasoning falls back to that reasoning instead of an empty string.
- User bubbles keep every word — only the think markers are dropped — so pasted markup can no longer swallow the bubble. New chat titles are built from the marker-free text.
- `useScrollSettler` re-asserts the scroll position across the settling window (animated while the send transition runs, instant afterwards). `recomputeBlankSpace` re-snaps whenever it moves the inset, a drag cancels the pin so the user is never fought, and a send in flight now outranks the keyboard-dismiss gate. A 300 ms fallback runs the pin even if `onContentSizeChange` never arrives.

## Test plan

- `yarn test` — 772 passing, incl. new `__tests__/messageText.test.ts` (15) and `__tests__/useScrollSettler.test.ts` (8, fake timers: correction order and animation mode, cancel-on-drag, restart mid-window, unmount cleanup).
- `yarn lint`, `yarn format:check`, `npx tsc --noEmit` clean for the touched files.
- Not verified on a device: the settler's root cause is derived from the app code and the keyboard-controller sources, not from a reproduced failure on hardware. Worth a manual pass — send several messages in a row, including from a scrolled-up position and with a multi-line input.

Closes #269


NorbertKlockiewicz (approved):
Worked as intended.

PR #291, opened 2026-08-10
fix(chat): update keyboard-controller past the scroll-worklet crash

kfaracik:
## What

Fixes the `SIGABRT` that closes the app when switching between model variants (Gemma 4 2B → VL). Five commits:

1. `5226d31` — bump `react-native-keyboard-controller` 1.21.1 → 1.22.3
2. `a185866` — await `loadModel()`, reset the shared values, unfreeze the scroll layout on model change
3. `79edd7a` — defer the load until the picker sheet has finished closing; add the native guard
4. `2dc205d` — replace the ad-hoc switch state with a state machine that cannot get stuck; rewrite the guard so it is visible
5. `2f8fcea` — revert the app-identity changes `a185866` had made for side-by-side installs

## Root cause

Crash reports off the reporter's iPhone 17 (`1.2.0 (3)`, three from 2026-08-10, one from 08-11) all carry the same stack:

```
abort() ← libc++abi terminate ← __cxa_throw
  ← hermesvm HermesRuntimeImpl::throwPendingError()
  ← worklets::WorkletRuntime::runSync / runSyncOnRuntime
  ← reanimated::UIEventHandler::process
  ← reanimated::UIEventHandlerRegistry::processEvent
  ← reanimated::ReanimatedModuleProxy::handleRawEvent
  ← React ScrollViewEventEmitter::onScrollEndDrag
  ← -[RCTScrollViewComponentView scrollViewWillEndDragging:…]
```

An uncaught JS error thrown inside a Reanimated worklet dispatched for a scroll view's drag-end. Nothing catches a `jsi::JSError` anywhere between `__cxa_throw` and `UIEventHandler::process`, so `terminate()` aborts the process — no red box, no error UI, the app just closes.

Two things it is **not**:

* **Not an OOM.** The `JetsamEvent` from the same session has the app alive and `frontmost` at 1.78 GB while *other* processes were killed.
* **Not the model load.** The same Gemma 4 VL build answered three vision prompts two minutes after the crashes.

In one report the JS thread sits in `worklets::runOnUISync`, blocked on the runtime lock the throwing worklet holds — the scroll view's worklet handlers were being registered/unregistered at the very moment the drag ended.

## The dependency bump alone does not fix it

`1.22.3` carries the two upstream fixes for that class of bug in `KeyboardChatScrollView` — [#1485](https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1485) (skip `scrollTo()` when the animated ref no longer resolves) and [#1579](https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1579) (unregister by stored view tag instead of re-resolving a detached fiber).

That was the original theory for this PR. **It is wrong on its own.** A release build with the bump *and* the deferral but *without* the native guard still crashes — reproduced twice on the reporter's iPhone 17.

That result is what justifies the patch. Upstream fixed one reason the worklet throws; the dispatch path still has no catch at all, so any other error there aborts the process just the same.

## The native guard

`patches/react-native-reanimated+4.3.1.patch` wraps the worklet call in `UIEventHandler::process`, which is the single point where Reanimated runs a worklet synchronously in response to a native event.

* **Narrow by design.** An exact-match allowlist of the five scroll events, and nothing else. Every other event class runs completely unchanged — one hash lookup, no try/catch. App Store crash reports are this app's only production signal, so widening the catch would trade that signal away for failure modes we have never seen.
* **The names are the ones that actually arrive.** Fabric emits `"scrollEndDrag"`, `EventEmitter::normalizeEventType` rewrites it to `"topScrollEndDrag"`, and Reanimated's `handleRawEvent` rewrites the prefix again to `"onScrollEndDrag"`. Matching the raw Fabric name would compile fine and silently guard nothing.
* **It logs where you can see it.** `worklets::PlatformLogger` → `os_log` on iOS, `__android_log_print` on Android, so a suppressed error is visible in Console.app / logcat *in release builds*. The previous revision of this patch used `fprintf(stderr)`, which release builds discard — a crash converted into an invisible failure.
* **Rate limited** to 10 reports per process launch. A worklet that throws once throws again on the next scroll frame, and this runs on the UI thread.
* **Per handler.** The catch is inside the loop over handlers, so one throwing worklet does not stop the others listening to the same event.

It suppresses rather than fixes: the symptom becomes a skipped scroll-animation frame instead of a process abort. It should be removed once the underlying throw is found and fixed upstream — the `jsi::JSError` message and stack it now logs are what that report will be built from.

Regenerating the patch needs `--exclude '(^|/)(\.cxx|build)/'`, otherwise `patch-package` sweeps in the local Android build output and the patch grows to 1102 files.

## The JS side

Loading a model unloads the previous one and hands a large amount of work to the native side. Doing that while the picker sheet is still animating away puts the message list's scroll view into mount/teardown while scroll events are still being dispatched — the race the stack above shows. So the load now waits for the sheet to finish closing.

`useModelSwitch` holds that as one state machine — `idle → awaitingDismissal → loading` — replacing four refs (`pendingModelRef`, `selectionPendingRef`, `currentSnapIndexRef`, `isModelSwitchingRef`) spread across two components. Every non-idle state has a guaranteed exit, so the header spinner and the blocked chat bar cannot outlive the switch:

* the sheet reopening releases `awaitingDismissal`, so an interrupted dismissal no longer strands the pick
* a 5s timeout releases it if `onDismiss` never arrives at all
* `loading` is released when the load settles, rejection included — the home screen's `handleSetModel` has no `try/catch` of its own, so an unconsumed rejection there would have left an unhandled promise on every failed switch
* the deferral frames are cancelled on unmount

`ModelSelectSheet` goes back to being a picker: it reports the tap and dismisses, and the caller owns the deferral. The chat screen and the home screen drive the same hook, so picking a model behaves the same in both.

While switching, the header shows the incoming model with a spinner and the chat bar keeps the draft and explains itself instead of silently dropping a send.

### Dropped from the earlier commits

`a185866` forced a full `ChatScreen` remount via a model-keyed `key`. That is removed. `Messages` takes no `model` prop and none of its refs are model-derived, so the remount reset nothing that needed resetting — it only threw away the scroll position and replayed the fade-in on every switch, while adding a mount/teardown of the very scroll view that crashes.

## Testing

* `yarn test` — 775 passed (8 new, covering the state machine's exit guarantees: interrupted dismissal, timeout, rejection, unmount, double-pick)
* `yarn lint` — 0 errors; `tsc` — no new errors
* Release build on a physical iPhone 17, verified at the binary level (`strings`) that the guard is actually linked in rather than served from stale DerivedData

## Not verified yet

* **The final build is still under test on the reporter's device.** What is confirmed is the negative result — without the guard it still crashes. Confirmation that the guard build survives the repro is pending.
* The exact JS error text is still unknown. An `.ips` does not carry the abort message. The `PlatformLogger` line added here is what will capture it on the next occurrence.


components/bottomSheets/ModelSelectSheet.tsx
```diff
@@ -21,23 +19,19 @@ import { Feedback } from '../../utils/Feedback';
...
   bottomSheetModalRef: RefObject<BottomSheetModal | null>;
-  selectModel: (model: Model) => void;
+  onModelPicked: (model: Model) => void;
   onSheetStateChange?: (isOpen: boolean) => void;
 }
 
 const ModelSelectSheet = ({
   bottomSheetModalRef,
-  selectModel,
+  onModelPicked,
```
NorbertKlockiewicz, line 28: It's not changed in the /components/benchmark/ModelSelector.tsx tho error is shown when trying to change the model there.

NorbertKlockiewicz (changes requested):
fix this one thing, besides seems to work

NorbertKlockiewicz (approved):

PR #292, opened 2026-08-11
fix: settings toggle, citation, model-name and navigation follow-ups after #289

kfaracik:
Five commits that were left stranded on `fix/attach-download-and-penalty` after [#289](https://github.com/software-mansion-labs/private-mind/pull/289) was squash-merged. Because the squash rewrote that branch's history, a PR straight from it would replay all 17 already-merged commits as a 51-file diff — so these are rebased onto `main` here, which makes the diff exactly the outstanding delta.

## What's in it

* **`fix(citations)`** — a decimal number no longer ends the cited sentence early, so `3.5` stops splitting a citation in half.
* **`feat(settings)`** — the response-speed stats under an assistant reply become optional. Adds `ThemedSwitch`, `SettingsSection` and `SettingsToggleRow`, and persists `showPerformanceMetrics` in the settings store.
* **`fix(chat)`** — reloading a chat's history with unchanged content no longer blanks the message list mid-scroll.
* **`fix(navigation)`** — leaving a model sheet `replace`s instead of `push`ing, so the back stack stops collecting duplicate model-hub entries.
* **`fix(drawer)`** — an interrupted collapse of the search overlay never reports `finished`, which left its `Modal` mounted and swallowing every touch underneath; it now also unmounts on a timer.
* **`refactor(ui)`** — review follow-up: the switch's track/thumb metrics and toggle duration move to `constants/switch.ts`, the drawer's unmount grace to `constants/drawer-layout.ts`, and the knob colour stops being a hard-coded `#FFFFFF` in favour of `bg.switchThumb` in both palettes.

## Testing

`yarn lint` (0 errors), `yarn test` (768 passed), `npx tsc --noEmit` clean on the touched files, plus the iPhone 17 Pro simulator for the settings toggle and the drawer overlay.


NorbertKlockiewicz (approved):
lgtm

PR #293, opened 2026-08-13
Release 1.2.1

kfaracik:


NorbertKlockiewicz (changes requested):
Onboarding and the check disagree. `getStartingModels`still recommends models by raw RAM tiers, but the new check uses 0.8 × RAM plus a bigger overhead formula. 

Result: on many devices, onboarding shows a recommended model whose download button is greyed out as "Incompatible". This happens in every tier — Gemma 4 VL for devices reporting 6–7GB, Qwen 3 1.7B and LFM 2.5 VL 1.6B on a 4GB iPhone, Qwen 0.6B on 2GB phones.

NorbertKlockiewicz (approved):

PR #294, opened 2026-08-13
fix(models): size the compatibility gate by download size, not parameters

kfaracik:
## Why

On a 6GB phone, downloading and loading **Gemma 4 VL - 2B** (4.0 GB on Android) silently kills the app. There is no crash dialog, no toast, no error state — the process is simply gone.

The compatibility gate that was supposed to prevent this sized models by **parameter count** instead of by **download size**: `2.0B params × 1.75 = 3.5 GB`, which fits under a 5.5 GiB device, so the model was reported compatible and the download button stayed enabled. The file on disk is 4.0 GB.

## Evidence from a physical Galaxy S20 FE (SM-G781B, 5.50 GiB RAM, Android 13)

The app is not crashing — the kernel low-memory killer reaps it while it is in the foreground:

```
08-13 16:02:51.628 I/lmkd    (  628): Reclaim 'com.swmansion.privatemind' (22410), uid 10448,
                                      oom_score_adj 0, state 2 to free 14800kB rss, 377076kB swap;
                                      reason: min2x watermark is breached even after kill
08-13 16:02:52.252 I/ActivityManager( 1256): Process com.swmansion.privatemind (pid 22410)
                                      has died: fg  TOP (659,280)
08-13 16:02:52.992 I/lmkd    (  628): 1(delay), 0(swap), 0(freelimit), 0(reentrymode)
                                      memory pressure events were skipped after a kill!
```

`oom_score_adj 0` is the foreground bucket — lmkd killed the app the user was looking at. There is no `Fatal signal`, no `SIGSEGV`/`SIGABRT` and no tombstone anywhere in the capture, which rules out a native or JS crash.

Reproduced twice, both times dead within ~10 s:

| scenario | app RSS | MemAvailable | outcome |
| --- | --- | --- | --- |
| switch from Gemma 4 – 2B to VL | 257 MB | 676 MB | killed ≤10 s |
| **cold start, no model loaded** | 285 → 398 → 91 MB | 2.34 GB → 636 MB | killed ≤10 s |

The cold-start run is the decisive one: RSS collapsing from 398 MB to 91 MB while `MemAvailable` drains is the kernel evicting pages of the memory-mapped `.pte`. The model does not fit regardless of what was loaded before it, so the model-switch path is not implicated.

`MemTotal` on this device is `5 763 304 kB` (5.50 GiB), and only 2.34 GB is available with the app stopped.

## Not a regression

The bug is identical in 1.2.0. `utils/modelCompatibility.ts`, `store/llmStore.ts` and `components/model-hub/ModelCard.tsx` are byte-identical between `0375de9` (the 1.2.0 release commit) and `main`, and `git log -S "isModelCompatible"` has never touched the model-loading path.

## What changed

`getModelMemoryRequirement` now derives the requirement from `model.modelSize` — the actual download size — instead of the parameter count:

```
requirement = modelSize × 1.3 + 0.5 GB
compatible  = requirement ≤ totalRAM × 0.8
```

`× 1.3` covers the KV cache and the runtime's working set on top of the mapped weights, `+ 0.5 GB` the JS/RN heap and tokenizer, and the `0.8` fraction leaves the system its share of a phone's RAM. The quantized-keyword heuristic is gone: the download size already reflects the quantization, so a filename no longer decides the verdict.

Calibrated against the tiers `getStartingModels` already defines, so the gate now agrees with what onboarding recommends:

| device | Gemma 4 VL (4.0 GB) | Gemma 4 (2.5 GB) | LFM 2.5 VL 1.6B (2.43 GB) | Qwen 3 1.7B (2.16 GB) |
| --- | --- | --- | --- | --- |
| 4 GB (3.70 GiB) | blocked | blocked | blocked | blocked |
| 6 GB (5.50 GiB) | **blocked** | allowed | allowed | allowed |
| 8 GB (7.50 GiB) | allowed | allowed | allowed | allowed |

On the S20 FE, Gemma 4 VL is the only model in the catalogue this blocks — precisely the one that OOMs.

`ModelCard` already disables the download button and greys the card when `isModelCompatible` is false, so no UI change was needed; the gate simply tells the truth now. The gate sits only on the `NotStarted` download button — a model that is already on disk keeps its delete button, so a user who downloaded Gemma 4 VL before this change can still reclaim the 4 GB.

## Declared minimums

The formula covers whatever the catalogue gains next, but for a model we have measured on hardware a heuristic is a weaker statement than a decision. `MODEL_MIN_RAM_GB` lets a model state its floor outright — when a model has an entry it decides, and everything else falls back to the size formula:

```ts
export const MODEL_MIN_RAM_GB: Record<string, number> = {
  'Gemma 4 VL - 2B': 7,
};
```

7 GB sits above what a 6GB handset reports (~5.5 GiB) and below an 8GB one (~7.5 GiB), which is the same boundary `getStartingModels` already draws. Gemma 4 VL is the only entry so far because it is the only model with a measurement behind it.

It is keyed by model name rather than a new database column, so the catalogue can declare requirements without a schema migration. An explicit entry wins in both directions: it can block a model whose file happens to be small, and it can clear one the formula is too pessimistic about.

## Scope

This closes the **download** path, which is what the report asked for. Two gaps remain and are deliberately out of scope here:

* A model that is **already downloaded** can still be picked in `ModelSelectSheet` and loaded — the sheet does not consult the gate, and neither does `llmStore.loadModelInstance`. On a device that already has Gemma 4 VL on disk, this PR does not stop the kill.
* `getStartingModels` keeps its own independent RAM tiers rather than deriving them from this gate.

Worth a follow-up that wires the gate into the load path so no route can bypass it.

## Testing

* `yarn test` — 803 passing, including regression tests that pin Gemma 4 VL blocked on a 6GB device and still available on an 8GB one, cover the 4GB and 6GB starting tiers, and check that a declared minimum overrides the formula in both directions. Fixtures are named for the RAM tier rather than for a handset.
* `yarn lint` — 0 errors.
* `npx tsc --noEmit` — clean for the touched files.
* Verified on a physical 6GB device with a release build: Gemma 4 VL shows an "Incompatible" badge with its download disabled, Gemma 4 - 2B stays available, and the delete button on the already-downloaded copy still works.

### DIFF
--- a/__tests__/defaultModels.test.ts
+++ b/__tests__/defaultModels.test.ts
@@ -7,6 +7,7 @@ import {
   getGenerationConfigForModel,
   getStartingModels,
 } from '../constants/default-models';
+import { isModelCompatibleWithRam } from '../utils/modelCompatibility';
 
 describe('getGenerationConfigForModel', () => {
   it('passes through the registry config without injecting a penalty', () => {
@@ -20,19 +21,26 @@ describe('getGenerationConfigForModel', () => {
 });
 
 describe('getStartingModels', () => {
-  it('returns low-end model suggestions below 4 GB RAM', () => {
+  it('always returns low-end model suggestions below 4 GB RAM', () => {
     expect(getStartingModels(3.99)).toEqual([
+      'Qwen 3 - 0.6B',
+      'LFM 2.5 VL - 450M',
+      'LFM 2.5 - 1.2B',
+    ]);
+    expect(getStartingModels(0)).toEqual([
       'Qwen 3 - 0.6B',
       'LFM 2.5 VL - 450M',
       'LFM 2.5 - 1.2B',
     ]);
   });
 
-  it('returns mid-range model suggestions from 4 GB to 6 GB RAM', () => {
+  it('replaces incompatible mid-range candidates with low-end fallbacks', () => {
+    // Qwen 3 - 1.7B and LFM 2.5 VL - 1.6B need more than 0.8 * 4GB once
+    // runtime overhead is applied, so they are replaced with weaker models.
     expect(getStartingModels(4)).toEqual([
-      'Qwen 3 - 1.7B',
       'LFM 2.5 - 1.2B',
-      'LFM 2.5 VL - 1.6B',
+      'Qwen 3 - 0.6B',
+      'LFM 2.5 VL - 450M',
     ]);
     expect(getStartingModels(6)).toEqual([
       'Qwen 3 - 1.7B',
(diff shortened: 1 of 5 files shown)


NorbertKlockiewicz (changes requested):
Onboarding and the check disagree. getStartingModelsstill recommends models by raw RAM tiers, but the new check uses 0.8 × RAM plus a bigger overhead formula.

Result: on many devices, onboarding shows a recommended model whose download button is greyed out as "Incompatible". This happens in every tier — Gemma 4 VL for devices reporting 6–7GB, Qwen 3 1.7B and LFM 2.5 VL 1.6B on a 4GB iPhone, Qwen 0.6B on 2GB phones.

NorbertKlockiewicz (approved):

PR #299, opened 2026-08-19
Dynamic model catalog from a remote manifest, DEFAULT_MODELS as fallback only

### LINKED ISSUE #298: Dynamically import the model catalog from HuggingFace — hardcoded DEFAULT_MODELS as fallback only
Supersedes #285. Part of the multilingual audit. Child of #287.

## Problem

The model catalog is a hardcoded array, `DEFAULT_MODELS` (`constants/default-models.ts:99`), with each entry's ExecuTorch sources wired in from a fixed `react-native-executorch` import list (`constants/default-models.ts:4-19`). Adding a model — e.g. Qwen 3.5, Qwen 3 - 4B, Phi-4-mini (#285) — currently means a code change and a release.

That means the catalog can only ever be as current as the last app release, and every new model export from `react-native-executorch` needs a PR + store rollout before users can see it.

## Proposal

Fetch the model catalog dynamically (e.g. from a HuggingFace org/space, or another versioned manifest source) at runtime, so new models can be added without an app release. Keep today's hardcoded `DEFAULT_MODELS` list as the **fallback** — used only when the dynamic fetch fails (offline, endpoint down, malformed response) — not as the primary source of truth.

Open questions this issue should resolve during implementation:
- Where the manifest lives and how it's authenticated/versioned.
- How compatibility gating (`isModelCompatibleWithRam`, `constants/default-models.ts:3`) and per-model generation/context profiles (`constants/model-profiles.ts`) apply to catalog entries that didn't exist at build time.
- Caching policy for the fetched catalog (staleness, offline-first launch).

## Relationship to #168

#168 is about a user pasting an arbitrary HuggingFace repo to import a one-off model — blocked because ExecuTorch export isn't a standardized HF format yet, per the maintainer comments there. This issue is narrower and unblocked by that: it's about the *curated* catalog we already control (the models software-mansion itself exports and lists in `DEFAULT_MODELS`) becoming remotely updatable instead of hardcoded, not about browsing arbitrary community repos.

kfaracik:
## Summary
- App startup fetches the model catalog from a versioned JSON manifest instead of only the hardcoded `DEFAULT_MODELS`, so new models can be added without an app release — validated with zod, cached in AsyncStorage for offline launches, falling back to `DEFAULT_MODELS` only when the fetch, the cache, and everything else has failed.
- Manifest is hosted on this repo's existing GitHub Pages site (`catalog/model-catalog.json` is the source of truth, published by CI on every push to `main` that touches the file).
- RAM compatibility gating and download blocking for incompatible devices already worked off `modelSize`/`isModelCompatibleWithRam` and needed no changes — it applies automatically to dynamically-fetched models too.

Closes #298

## Test plan
- [x] `yarn lint` / `yarn format:check` / `yarn test` (815/815) all green
- [x] `catalog/model-catalog.json` validated against the zod schema in CI (`modelCatalogFile.test.ts`)
- [ ] Manual on-device verification of the model-hub list (blocked in this session by an unrelated stale Metro bundle in the local dev environment — not re-attempted after the HF/GitHub back-and-forth)


kfaracik:
# Model catalog: Hugging Face setup

> **Status:** not active yet. The app currently fetches the catalog from
> GitHub Pages (`constants/modelCatalog.ts`), which needs no setup and works
> today. This doc is the checklist for switching to Hugging Face hosting
> once the org-side setup below is done — see "Switching over" at the
> bottom for the exact code change.

The app fetches its model catalog at runtime from a JSON manifest
(`constants/modelCatalog.ts`). The manifest's source of truth is
`catalog/model-catalog.json` in this repo — editing it and merging to `main`
is what updates the catalog for already-installed apps, with no app release.

Once switched over, CI will publish that file to Hugging Face automatically
(`.github/workflows/test.yml`, job `publish-catalog`). Before that first run,
someone with access to the `software-mansion` Hugging Face org needs to do
the following one-time setup:

## 1. Create the dataset repo

On huggingface.co, create a new **dataset** repository under the
`software-mansion` org:

- Name: `private-mind-model-catalog`
- Visibility: **public** (the app fetches it with a plain, unauthenticated
  `resolve` URL — a private repo would need a bearer token the app doesn't
  send)

This gives the repo `https://huggingface.co/datasets/software-mansion/private-mind-model-catalog`.

## 2. Create an access token

Generate the token from a **personal** Hugging Face account that has write
access to the `software-mansion` org (Hugging Face access tokens are tied to
a personal account, not an org, even when used to write to an org-owned
repo). Settings → Access Tokens → New token:

- Type: fine-grained
- Permissions: write access, scoped to only the
  `private-mind-model-catalog` repo (not the whole org)

## 3. Add the token as a GitHub secret

In this GitHub repo: Settings → Secrets and variables → Actions → New
repository secret.

- Name: `HF_TOKEN`
- Value: the token from step 2

## 4. Verify

Merge a change to `catalog/model-catalog.json` on `main`, let the
`publish-catalog` job run, then confirm the manifest is live:

```
curl https://huggingface.co/datasets/software-mansion/private-mind-model-catalog/resolve/main/model-catalog.json
```

If the URL above ever changes (different org, repo name, or revision), update
the default in `constants/modelCatalog.ts` to match — or override it per
build via the `EXPO_PUBLIC_MODEL_CATALOG_URL` env var without touching code.

## 5. Switching over

Once steps 1–4 are done:

- `constants/modelCatalog.ts` — change the `MODEL_CATALOG_URL` default to
  `https://huggingface.co/datasets/software-mansion/private-mind-model-catalog/resolve/main/model-catalog.json`.
- `.github/workflows/test.yml`, job `publish-catalog` — replace the
  `Stage the manifest` / `Publish to gh-pages` steps with a single step that
  runs `pip install -U "huggingface_hub[cli]"` then:

  ```
  hf upload software-mansion/private-mind-model-catalog \
    catalog/model-catalog.json model-catalog.json \
    --repo-type dataset \
    --token "$HF_TOKEN" \
    --commit-message "chore: publish model catalog"
  ```

  guarded by the same `steps.catalog-changed.outputs.changed == 'true'`
  condition already on those steps, with `env: HF_TOKEN: ${{ secrets.HF_TOKEN }}`
  on the step. Drop the job's `permissions: contents: write` block — it's
  only needed for pushing to this repo's own `gh-pages` branch.

NorbertKlockiewicz (approved):

PR #304, opened 2026-08-30
fix(chat): stop small models looping without breaking grounded answers

kfaracik:
Two changes to how repetition is handled, replacing the old global 1.1 penalty (which was the main cause of wrong numbers and script drift in grounded RAG answers — see #255) without reintroducing the loops it masked.

**Sampler patch (`patches/react-native-executorch+0.9.2.patch`)** — the upstream sampler penalizes every prompt token per occurrence, so the multiplicative form compounds exponentially for anything quoted repeatedly in the prompt; pasted context flattened the distribution for exactly the tokens the answer needed. The patched sampler applies the penalty only over the last 256 tokens and once per distinct token.

**Scoped overrides (`constants/default-models.ts`)** — `repetitionPenalty` stays off registry-wide and is overridden at a conservative **1.05** only for the two models observed degenerating into verbatim loops on ordinary prompts with it off: Qwen 2.5 0.5B and Bielik 1.5B.

Tests cover the registry pass-through default and both overrides (8/8 green).


kfaracik:
Converting this to an issue: #375.

The code here still holds up — the sampler patch is the right diagnosis — but the PR cannot be merged as written: it branched 2026-08-30 and `constants/default-models.ts` has since been restructured around model profiles, so its bespoke `GENERATION_CONFIG_OVERRIDES_BY_MODEL_PATH` map duplicates what `PROFILE_BY_MODEL` now does, and the patch is pinned to 0.9.2 so the #344 bump would silently drop it.

Since opening it we also filed the root cause upstream, where it belongs: https://github.com/software-mansion/react-native-executorch/issues/1483 — the whole-prompt, once-per-occurrence penalty, plus a request for DRY, XTC and locally typical sampling. That changes the calculus on whether we carry a local patch at all, which is the decision #375 now tracks.

Branch `fix/repetition-penalty-window` is kept.

PR #311, opened 2026-09-04
fix(chat): cut a reply that collapsed into a repeated clause (#255)

kfaracik:
## Problem

#289 reverted the blanket `repetitionPenalty` from #258: ExecuTorch applies the penalty over the whole prompt with no sliding window, so retrieved context suppressed its own words and grounded answers turned to gibberish (2/2 in a controlled A/B). That reopened #255 — a small quantized model can still collapse into `czapki, czapki, czapki…` and run until the token limit.

## Fix

A guard on the finished text instead of on sampling: `truncateAtRepeatedClause` cuts at the first repetition it can prove (a cycle of up to six clauses repeated twice, one clause three times, a line twice, a word four times in a row, a phrase of up to five words three times). If the cut would eat the whole reply, the first sentence is kept. It reads repetition, not words, so it holds in every language.

Wired at the one place `sendChatMessage` finalizes the reply on `main`.

## Evidence

Qwen 3 1.7B, 259 stored answers: three of the four longest replies were loops this cuts. The remaining cost is time — the loop has already run before it is cut; stopping generation mid-stream on a detected cycle is the follow-up.

## Provenance

The same guard has been running on the `web-search-compact` branch since 2026-08-21 (`a3f64af`). It is extracted here so it can be reviewed and land on its own; once this merges, that branch's diff against `main` no longer carries it. Device testing of `web-search-compact` treats loop-guard behaviour as out of scope for that branch.

## Tests

- `__tests__/loopDetection.test.ts`, `__tests__/loopTruncationKeepsAnswer.test.ts` — the guard's rules and the "never destroy the whole answer" invariant
- `__tests__/llmStore.test.ts` — a reply that repeats one clause four times is persisted with one copy (red without the wiring)

### DIFF
--- a/__tests__/llmStore.test.ts
+++ b/__tests__/llmStore.test.ts
@@ -447,6 +447,27 @@ describe('sendChatMessage', () => {
     expect(mockPersistMessage).not.toHaveBeenCalled();
   });
 
+  it('cuts a reply that collapsed into a repeated clause before persisting it (#255)', async () => {
+    mockInstance.generate.mockResolvedValue(
+      'Czapka kosztuje 45 zł. Czapka kosztuje 45 zł. Czapka kosztuje 45 zł. Czapka kosztuje 45 zł.'
+    );
+    useLLMStore.setState({
+      model: baseModel,
+      activeChatId: 1,
+      activeChatMessages: [],
+    });
+
+    await useLLMStore
+      .getState()
+      .sendChatMessage('ile kosztuje czapka?', 1, noSources, settings);
+
+    const persisted = mockPersistMessage.mock.calls
+      .map((call) => call[1] as { role: string; content: string })
+      .find((message) => message.role === 'assistant');
+    expect(persisted?.content).toContain('45 zł');
+    expect(persisted?.content.split('Czapka kosztuje').length - 1).toBe(1);
+  });
+
   it('persists user message and assistant response', async () => {
     useLLMStore.setState({
       model: baseModel,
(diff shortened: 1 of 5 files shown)


kfaracik:
Superseded — this landed on `main` already, and in a better shape than this branch carries.

The guard rode in with the web-search work: `9a15bfd feat(web): on-device web search for chat (#320)`, refined in `befb4a5 (#334)`. This branch is based on `b3aa96f` (Release 1.2.1), two releases back, which is where the conflict comes from.

File by file, against `main` today:

- `utils/loopDetection.ts` — superset: same rules, plus an extracted `loopCuts` and a new `isRepetitionFromTheStart`
- `__tests__/loopDetection.test.ts` — byte-identical
- `__tests__/loopTruncationKeepsAnswer.test.ts` — identical bar the comments
- `store/llmStore.ts` — wired in four places through `tidyVisibleAnswer`, which also skips think blocks; this branch wires one
- `__tests__/llmStore.test.ts` — this one case is absent by name, but the path is covered more broadly (`deviceAnswerCorpus.test.ts`, think-block loops, the continuation nudge, the iPhone SE case)

`main` also closes a gap this branch leaves open: a reply that is a loop from its first character used to survive untouched, because cutting at offset 0 falls through to the keep-the-answer salvage. `isRepetitionFromTheStart` is used in `carriesAnswer` so the bubble is dropped instead.

Resolving the conflict in this branch's favour would revert all of that, so closing rather than rebasing. #255 stays open for what is genuinely left: emoji walls are not caught (the alphanumeric guard skips them), and the loop still runs to the token cap before it is cut — stopping mid-stream is the follow-up.

kfaracik:
Closing: superseded by #320 / #334 as detailed above. The branch stays up in case anything needs pulling from it.

PR #314, opened 2026-09-09
feat(web): default web search off for a model too small to have been measured

kfaracik:
## What

Two changes, both about the same gap: nothing stopped web search from being offered on a model that cannot use it.

1. **A capability floor for unmeasured models.** `PROFILE_BY_MODEL` decides readiness only for the models the cross-model pass actually ran. Everything else — a future catalogue entry, a remote-manifest entry, a model the user imports — falls through to `DEFAULT_PROFILE`, where `webSearchReady` is `true`. `WEB_SEARCH_MIN_PARAMETERS_B = 0.7` gates a model that is both below the floor **and** has no explicit `webSearchReady`. An explicit entry still wins, so a small model that grounds well can be admitted by name; a model of unknown size is left enabled rather than guessed at.

2. **The toggle now reads what the send will actually do.** `ChatScreen` passes `chatSettings.webSearchEnabled && webSearchUsable` to `ChatBar` instead of the raw setting. The stored setting is untouched, so it comes back when the user returns to a capable model.

## Why the number, and how much weight it carries

Parameter count is a proxy for the property we actually gate on, and not a good one. Two independent facts say so.

**The measurement.** A 56-turn pass over the English protocol scores the sub-1B group at 6/10 correct against 27/46 for everything at 1B and above — no difference. The best result relative to the context it was handed belongs to `Qwen 3 - 0.6B` (4/4); the worst belongs to `Gemma 4 - 2B` (0/2), a model 4.4x its size.

**The gate list is not sorted by size.** If size were what we gate on, it would be:

| model | parameters | gated | below the 0.7 floor |
|---|---|---|---|
| `LFM 2.5 VL - 450M` | 0.45 | yes | yes |
| `Qwen 2.5 - 0.5B` | 0.49 | yes | yes |
| `Qwen 2.5 - 1.5B` | 1.54 | yes | no |
| `Qwen 2.5 - 3B` | 3.09 | yes | no |
| `Qwen 3 - 0.6B` | 0.75 | no | no |
| `LFM 2.5 - 1.2B` | 1.20 | no | no |

Four models are gated; only two of them are below the floor. `Qwen 2.5 - 3B` is gated at 3.09B while `Qwen 3 - 0.6B` runs at 0.75B. What the four gated names have in common is an entry in `WEB_ANSWER_EVIDENCE`, not a size — `Qwen 2.5 - 0.5B` answered 33% correctly with retrieval at 100%, while `Qwen 3 - 0.6B` answered 71% on the same corpus.

So the floor is a default for the case where we have no evidence at all, not a claim that size predicts grounding.

## Effect today: none

No model in the app reaches the new branch.

- No catalogue model is below 0.7 without an explicit entry. The only candidate by name, `Qwen 3 - 0.6B`, is `parameters: 0.75` in `constants/default-models.ts` — above the floor.
- A user-imported model cannot reach it either: `app/(modals)/add-local-model.tsx` `handleSave` sets `modelName`, `source`, the three paths and `modelSize`, and never sets `parameters`. The floor requires `model.parameters !== undefined`.

The floor takes effect only for a future catalogue entry or a remote-manifest model from #299. Change 2, by contrast, changes behaviour now.

## What change 2 fixes

On `cr/phase1-security`, `ChatScreen` passes the raw setting. Switch from a capable model to a gated one with web search on and the toggle keeps showing on. The `modelSwitching` guard in `ChatBar.handleSend` only closes the loading window; after loading, the send goes through and the search is dropped silently in `useSendChatMessage` where `isWebSearchReady` enters the conjunction. The user sees an on toggle, sends, and gets an ungrounded answer with no explanation.

## Verified on a device

Samsung S20 FE (SM-G781B, Android 13), debug build over Metro, `projectRoot` confirmed as this worktree.

- Bielik v3.0, web search on → switch to `Qwen 2.5 - 0.5B` → wait for the load to finish → the toggle shows off **without being touched**.
- Tapping it then shows `This model cannot use web results reliably — pick a larger one` instead of enabling a search that would not run.
- Switching back to a capable model restores the toggle to on, so the setting was preserved rather than cleared.

The toggle reflects the previous model until `loadModel` resolves, because `selectModel` is awaited after it; sends are blocked by `modelSwitching` for that whole window.

## Tests

`__tests__/modelProfiles.test.ts` — an uncatalogued model below the floor is gated; at the floor is not; of unknown size is not; `Qwen 3 - 0.6B` stays above the floor and enabled; the catalogued gate list is exactly the four evidenced names; no catalogued model is gated without an entry in `WEB_ANSWER_EVIDENCE` or `PLANNER_EVIDENCE`.

`__tests__/chatScreenWebSearchState.test.ts` — the toggle reads off after moving to a gated model; reads on for a capable one; explains itself instead of clearing the setting; keeps the setting so returning restores it; still turns off on demand.

`tsc --noEmit` clean; `jest` green; eslint clean.

### DIFF
--- a/__tests__/chatScreenWebSearchState.test.ts
+++ b/__tests__/chatScreenWebSearchState.test.ts
@@ -1,0 +1,99 @@
+import { renderHook } from '@testing-library/react-native';
+
+const mockToast = jest.fn();
+jest.mock('react-native-toast-message', () => ({
+  show: (...args: unknown[]) => mockToast(...args),
+}));
+jest.mock('../store/chatStore', () => ({
+  useChatStore: () => ({
+    phantomChat: null,
+    setPhantomChatSettings: jest.fn(),
+  }),
+}));
+jest.mock('../store/llmStore', () => ({
+  useLLMStore: () => ({ model: null, loadModel: jest.fn() }),
+}));
+jest.mock('../store/modelStore', () => ({
+  useModelStore: () => ({ getModelById: jest.fn() }),
+}));
+jest.mock('../database/chatRepository', () => ({
+  checkIfChatExists: jest.fn(),
+  setChatSettings: jest.fn(),
+}));
+jest.mock('../utils/modelCompatibility', () => ({
+  hasMemoryForWebSearch: () => true,
+}));
+
+import { useChatScreenActions } from '../components/chat-screen/useChatScreenActions';
+import type { Model } from '../database/modelRepository';
+
+const modelNamed = (modelName: string, parameters: number): Model =>
+  ({
+    id: 1,
+    modelName,
+    parameters,
+    source: 'built-in',
+    isDownloaded: true,
+    modelPath: '',
+    tokenizerPath: '',
+    tokenizerConfigPath: '',
+  }) as Model;
+
+const setSetting = jest.fn();
+
+const actionsFor = (model: Model, webSearchEnabled: boolean) =>
+  renderHook(() =>
+    useChatScreenActions({
+      chatId: 1,
+      chat: undefined,
+      model,
+      chatSettings: {
+        systemPrompt: '',
+        thinkingEnabled: false,
+        webSearchEnabled,
+      },
+      setSetting,
+      db: {} as never,
+      inputRef: { current: null },
+    })
+  ).result.current;
+
+describe('web search state follows the model that has to run it', () => {
+  beforeEach(() => {
+    setSetting.mockClear();
+    mockToast.mockClear();
+  });
+
+  it('reads as off once the chat moves to a model that cannot search', () => {
+    const actions = actionsFor(modelNamed('Qwen 2.5 - 0.5B', 0.49), true);
+    expect(actions.webSearchEnabled).toBe(false);
+  });
+
+  it('reads as on for a model that can', () => {
+    const actions = actionsFor(modelNamed('Qwen 3 - 1.7B', 2.03), true);
+    expect(actions.webSearchEnabled).toBe(true);
+  });
+
+  it('explains itself instead of silently clearing the setting', () => {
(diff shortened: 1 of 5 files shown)


kfaracik:
Dołożyłem drugą połowę tej samej reguły, od strony "już włączone" — c6 potwierdziło, że nic w `ChatBar` nie jest w toku.

**`fix(chat): show web search as off when the model cannot run it`**

`handleWebSearchToggle` strzeże wyłącznie akcji włączenia (`!chatSettings.webSearchEnabled && !isWebSearchReady(model)`), więc przełączenie czatu na model z gate'em nie rewaliduje niczego: zapisana flaga zostaje na `true`, przełącznik dalej świeci, a `useSendChatMessage` po cichu pomija wyszukiwanie i tłumaczy się toastem dopiero przy wysyłce. Kontrolka mówiła co innego niż tura.

Przełącznik czyta teraz stan efektywny — zapisana flaga **oraz** model zdolny wyszukiwać na tym urządzeniu. Tapnięcie przy modelu z gate'em pokazuje ten sam komunikat co próba włączenia, zamiast po cichu kasować ustawienie, którego użytkownik nie zmieniał. Zapisana flaga zostaje nietknięta, więc powrót do modelu zdolnego przywraca web search tam, gdzie użytkownik go zostawił.

Pięć testów w `__tests__/chatScreenWebSearchState.test.ts`: czyta się jako wyłączony po przejściu na model z gate'em, jako włączony na modelu zdolnym, tłumaczy się zamiast czyścić ustawienie, zachowuje ustawienie na powrót, i nadal daje się normalnie wyłączyć.

Razem: `tsc` czysty, **129 zestawów / 2265 testów** zielone, eslint bez błędów.

NorbertKlockiewicz (approved):

PR #315, opened 2026-09-09
fix(web): drop a digest that is the digest prompt's own example

kfaracik:
## What

`DIGEST_SYSTEM_PROMPT` shows the model what a topic phrase should look like by spelling one out: `parzenie kawy w kawiarce, stopien zmielenia`. Small models hand that example back as their answer. It is then stored as the chat's digest and framed into the planner prompt on every later turn.

A digest whose whole text appears in the prompt it was generated from is now discarded at the source, and the turn falls back to the question.

## Evidence

From the cross-model pass on the Pixel, all three in an **English** conversation whose previous turn was "What is the weather in Warsaw today?":

| model | `sourceQuery` recorded in the DB |
| --- | --- |
| Qwen 3 - 0.6B | `What about tomorrow? parzenie kawy w kawiarce, stopien zmielenia` |
| Bielik - v3.0 | `weather tomorrow Warsaw parzenie kawy w kawiarce, stopien zmielenia` |
| LFM 2.5 - 1.2B | same suffix; the answer then attributed a fabricated "rain today" to a source that was never read |

Qwen 3 - 0.6B's turn returned five live Polish coffee-brewing pages and cited one of them as a weather source. The compatibility round reads this as a cache or embedding collision; it is neither.

## The stored digest, on a second device

The evidence above is the digest leaking into a query. Pulling `executorch.db` off a Samsung S20 FE shows the value itself, sitting in `chatSettings.digest`:

| chat | digest |
| --- | --- |
| 31 | `parzenie kawy w kawiarce, stopien zmielenia` |
| 32 | `parzenie kawy w kawiarce, stopien zmielenia` |
| 33 | `gold price` |

Chat 31 is "Ile kosztuje Samsung Galaxy S25 w Polsce?" → "And how much is it in Germany?" → "Wie ist das Wetter heute in Berlin?" → "What did I ask about first?". Chat 32 opens the same way. Neither mentions coffee, grinding, or a moka pot anywhere. Chat 33 is a turn run on this branch — "What is the current price of gold per ounce in USD?" — and its digest is the topic, not the example.

Chat 31's last answer reads `The source that you are most likely to be referring to is thoughtcatalog.com: "400+ First Date Questions…"`. That is the kind of retrieval drift a poisoned digest would produce on a referent question, but no build writes the digest to a trace: `recordWebSearchTrace` stores `planQueries`, the planner's output, not its input. Recorded as consistent with the hypothesis, not as proof of it.

## Why the query path was not enough

`carryReferentIntoQuery` already refuses a digest that shares no content stems with the conversation, so the coffee phrase no longer reaches a query. The digest itself was never filtered: it still goes into `buildConversation` as `Conversation summary so far: …`, into `topicAnchorer`, and into `languageReferenceFor`. This closes it at the point where the bad value is created rather than at one of its readers.

## Why this shape and not the obvious one

The first attempt required a digest to share vocabulary with the exchange it summarises. It failed an existing test, and the test was right: a model may summarise a Polish exchange in English, and that digest is still good. Comparing the digest against the prompt instead has no such failure mode, and it generalises — any example or instruction fragment we ever put in that prompt is covered, in any language.

`looksLikePromptEcho` sits next to `looksLikeAnswerEcho`, which drops a digest that copies the model's own answer. Same defect, two sources; the two predicates read as a pair.

## Tests

Six cases in `__tests__/conversationDigest.test.ts`: the example is recognised, an instruction fragment copied verbatim is recognised, a real topic is left alone, a single word is never judged, the stored digest falls back to the question, and a topic the model actually wrote survives. The pre-existing case that keeps an English digest of a Polish exchange still passes.

`tsc --noEmit` clean; `jest` 132 suites / 2297 tests green; eslint no errors.

## Not in this PR

Removing the literal example from the prompt would delete the cause outright rather than filter its effect, which is the better shape. It also changes format compliance for every model and needs its own measurement, so it is left separate.

A third way into the digest stays open, and this PR does not narrow it: when the visible text is empty, `visibleDigestText` falls back to an **unterminated** `think` block and returns the model's internal reasoning as the topic, provided it fits `DIGEST_MAX_CHARS`. That is P1-10 in the code-review plan. The guard added here rejects a digest that copies the prompt; it does not reject one that copies the model's scratchpad.

### DIFF
--- a/__tests__/conversationDigest.test.ts
+++ b/__tests__/conversationDigest.test.ts
@@ -1,7 +1,7 @@
 import {
   buildDigestPrompt,
   DIGEST_MAX_CHARS,
-  echoesItsOwnPrompt,
+  looksLikePromptEcho,
   looksLikeAnswerEcho,
   stripMetaFrame,
   updateConversationDigest,
@@ -322,19 +322,19 @@ describe('a small model can hand back the prompt it was given', () => {
   const EXAMPLE = 'parzenie kawy w kawiarce, stopien zmielenia';
 
   it('recognises the example the prompt itself spells out', () => {
-    expect(echoesItsOwnPrompt(EXAMPLE)).toBe(true);
+    expect(looksLikePromptEcho(EXAMPLE)).toBe(true);
   });
 
   it('recognises a fragment of the instructions copied out verbatim', () => {
-    expect(echoesItsOwnPrompt('Output ONLY that phrase')).toBe(true);
+    expect(looksLikePromptEcho('Output ONLY that phrase')).toBe(true);
   });
 
   it('leaves a real topic alone', () => {
-    expect(echoesItsOwnPrompt('Warsaw weather tomorrow')).toBe(false);
+    expect(looksLikePromptEcho('Warsaw weather tomorrow')).toBe(false);
   });
 
   it('judges nothing on a single word, which could land anywhere', () => {
-    expect(echoesItsOwnPrompt('phrase')).toBe(false);
+    expect(looksLikePromptEcho('phrase')).toBe(false);
   });
 
   it('stores the question instead of the example it was shown', async () => {
--- a/__tests__/conversationDigest.test.ts
+++ b/__tests__/conversationDigest.test.ts
@@ -1,6 +1,7 @@
 import {
   buildDigestPrompt,
   DIGEST_MAX_CHARS,
+  echoesItsOwnPrompt,
   looksLikeAnswerEcho,
   stripMetaFrame,
   updateConversationDigest,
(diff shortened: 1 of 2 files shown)


kfaracik:
Two additions from the parallel session, both worth having on the record.

**This already broke a real turn once, before the cross-model pass.** `docs/WEB_SEARCH_RAG_STATUS.md:443-451` describes the same leak on device: the digest came back as the prompt's example, degraded into `"zaporanie kawy w kawiarce"` (25 chars, Polish, query-shaped), the trace showed `Searching "A ile się ja parzy? zaporanie kawy w kawia…"`, and retrieval moved to `coffeepolska.pl` and `inpostfresh.pl`. So this is a repeat, not a first sighting.

**Scope, stated precisely.** 184 recorded traces were swept for stems of the example (`kawiar`, `zmielen`, `parzeni`, `kawy`, `kawe`) in `planQueries` and `retrievalQuery`, skipping traces whose question was genuinely about coffee. **Zero hits.** That confirms `digestDescribesConversation` holds the query path empirically, and it means this PR is not fixing that path — it is fixing the three unguarded readers:

| `buildSearchQuery.ts` | consumer | guarded |
| --- | --- | --- |
| 561-562 | `carryReferentIntoQuery` | yes |
| 624-625 | `buildConversation` → `Conversation summary so far: …` | no |
| 711 | `topicAnchorer` | no |
| 761 | `languageReferenceFor` | no |

One guarded reader and three unguarded ones is the reason to discard the value where it is created. Guarding each reader instead would mean three gates today and a fourth hole the next time someone adds a consumer.

**A gap this does not close, and cannot.** A digest that only *frames* the planner prompt is invisible in the traces by construction — `WebSearchTrace` records `planQueries`, the planner's output, not the digest or the prompt it was built from. So there is proof it no longer reaches a query verbatim, and no evidence either way about its effect as framing. Adding `digest` to `WebSearchTrace` in `utils/web/searchTrace.ts` would let the next device round measure it directly. Deliberately not done here: that file is under review right now, and editing it underneath would discard someone's progress. Filed as a follow-up.

NorbertKlockiewicz (approved):

Issue #252, opened 2026-07-20
Update Onboarding Carousel


kfaracik:
- [ ] Unify UI: Ensure consistent design (margins, typography, alignment) across all current slides.
- [ ] Add Web Search Slide: Create a new slide to introduce the Web Search feature.
- [ ] Update Legal Terms: Revise the Terms of Service (ToS) and Privacy Policy to ensure maximum legal compliance regarding the new web search feature.
- [ ] Communicate device-dependent performance: Add a slide (or a section on the model-selection/onboarding step) making it clear that on-device model speed and capability **depend heavily on the device** (RAM / chipset / GPU). Illustrate it with a chart or other visual rather than plain text — e.g. tokens/sec or time-to-first-token across low / mid / high-RAM tiers, or a "device RAM → recommended model" map. This sets expectations up front (why the same model is snappy on one phone and slow/unavailable on another) and reduces confusion before the user downloads a model.

  Notes for the visual:
  - Tie it to the existing device tiers (`utils/voice/deviceTier.ts`, `constants/default-models.ts` `getStartingModels`): `≤4 GB → low`, `≤6 GB → mid`, `>6 GB → high`.
  - Numbers should come from real on-device measurements (benchmark screen already captures tok/s + peak memory), not estimates.

- [ ] Fix the struck-through clock on the first slide: the status bar of `step_chat` shows **9:41 with a strikethrough line drawn over it**. The line is baked into the PNG (present at @1x/@2x/@3x), not rendered by the app — it looks like a leftover annotation exported from the design file. Visible on every device; on iPhone SE it is `assets/onboarding/step_chat@2x.png`.

  Notes:
  - Affected files: `assets/onboarding/step_chat.png`, `step_chat@2x.png`, `step_chat@3x.png` (used at `app/(modals)/onboarding.tsx`, first slide).
  - Only `step_chat` is affected — `step_voice` shows the same 9:41 without the line, `step_sources`/`step_models` show real capture times (11:49 / 11:54).
  - The `step_chat` status bar is drawn vectorially, not captured (crisp edges vs. the rescaled, slightly soft "11:54" in `step_models`), so the fix is to re-export or repaint that status bar rather than re-crop a screenshot.

PR #317, opened 2026-09-10
Onboarding carousel: add the Web Search slide and unify slide layout

kfaracik:
Draft — two of the five boxes in #252 are done, the rest is listed below.

Closes nothing yet; refs #252.

## What is in the branch

**A Web Search slide**, fourth of five, between "AI RAG" and "Speech to text":

> **Web search** — Search the web in chat
> *Let the model look things up online for fresh, sourced answers.*

with `step_websearch.png` at @1x/@2x/@3x.

**One layout rule for every slide.** The per-step `alignment: 'center' | 'bottom'` field is gone. Every image is now bottom-aligned above the panel and drawn at a fixed 300×618 with `resizeMode="contain"`, instead of two slides centring and three sitting on the panel edge at their intrinsic size.

## Still to do

- [ ] **`useThemedStyles` was replaced by the boilerplate it exists to remove.** The branch swaps `const { styles, theme } = useThemedStyles(createStyles)` for a hand-rolled `useTheme()` + `useMemo(() => createStyles(theme), [theme])`. `CLAUDE.md` names that pattern as the thing not to write, and `main` already used the hook here, so this is a regression rather than a new file getting it wrong. Likely an accidental revert to an older copy of the file. **This should go back before the PR leaves draft.**
- [ ] **Check 300×618 on a small screen.** The size is hardcoded while the wrapper is `SCREEN_HEIGHT` tall and offset by the measured panel height. `contain` will letterbox rather than overflow, but an iPhone SE and a tall Android need a look before this is called done.
- [ ] **Legal terms** (#252) — ToS and Privacy Policy still say nothing about web search.
- [ ] **Device-dependent performance slide** (#252) — not started. The issue asks for a visual tied to the existing tiers (`utils/voice/deviceTier.ts`, `getStartingModels`: ≤4 GB low, ≤6 GB mid, >6 GB high) with numbers from real benchmark captures, not estimates.
- [ ] **The struck-through clock on `step_chat`** (#252) — the 9:41 status bar still carries a baked-in strikethrough at all three densities. Untouched here; it needs a re-export, not a re-crop.
- [ ] Copy review on the new slide's three strings.

## Checks

`jest` 36 suites / 433 tests green. `eslint` on the changed file: 0 errors, 3 warnings, all pre-existing on `main` (`pageX` / `pageY` / shadowed `height` in the `measure` callback at what is now line 261).

`tsc --noEmit` reports 31 errors, every one of them in `__tests__/` and every one inherited from `main` — the branch is a single commit touching `onboarding.tsx` and three PNGs, so it introduces none of them.

No device round yet: the slide has not been seen running on a phone.


kfaracik:
Superseded by #337, which reworks the whole carousel (swipe, pagination, skip, reduced motion) and carries this branch's `step_websearch.png` assets unchanged — byte-identical at all three densities.

Worth recording before this closes: the `useThemedStyles` checkbox in the description was not a regression. This branch is based on 21 July, 24 commits behind `main`, so it predates #260 where the hook landed — the file simply still looked like that. On a current base the point disappears.

The unified-layout idea did not survive a device check either: with one anchor for every slide, either the document chip and the voice waveform fall under the panel, or the web search trace and the chat header get clipped off the top. #337 keeps the anchor per slide for that reason.

Closing as superseded; #252 stays open for legal terms, the device-performance visual and the struck-through clock in `step_chat`.

PR #321, opened 2026-09-11
chore(release): 1.3.0

kfaracik:
Version bump only. One commit, three files, no source changes.

`main` still declares `1.2.1`, the version already on the stores, so a build cut today would ship web search under a released version string. Minor rather than patch, since web search is a new user-facing feature.

All three locations from `docs/RELEASE.md`:

- `app.json` — `1.2.1` → `1.3.0`
- `android/app/build.gradle` — `versionCode` 68 → 69, `versionName` `1.2.1` → `1.3.0`
- `ios/PrivateMind.xcodeproj/project.pbxproj` — `MARKETING_VERSION` `1.2.1` → `1.3.0`, `CURRENT_PROJECT_VERSION` 4 → 5

The pod-lockfile fix that previously shared this branch now lives in #322. The two touch disjoint files and can merge in either order.

## This does not make `main` releasable

Still outstanding, and not addressed here:

- **The `cr/phase1-security` work is not on `main`.** 121 files, +6991/−1352 — the code-review fixes, the ECB conversion resolver, the capability floor, the digest fix and the URL-scheme check. `#320` merged as a squash, so the branch cannot be merged directly (72 conflicting files); the content has to be ported onto `main` as a tree copy instead.
- **Three call sites open a source URL with no scheme check** — `SourceRow.tsx:67`, `WebSearchBlock.tsx:103`, `DominantSourceBadge.tsx:19`. The markdown link handler in `MessageItem.tsx:128` is guarded; these are not.
- **Three web-search defects found on a physical Pixel 10** against the shipped `main` code: conflicting sources on the same date send the model into a reasoning loop that ends with no answer and no error; offline answers state a fabricated figure and cannot carry a grounding caveat, because caveats are disabled when the context is empty; a currency-conversion follow-up returns the original currency despite retrieval finding the right pages.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -99,8 +99,8 @@ android {
         applicationId 'com.swmansion.privatemind'
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 68
-        versionName "1.2.1"
+        versionCode 69
+        versionName "1.3.0"
 
         buildConfigField "String", "REACT_NATIVE_RELEASE_LEVEL", "\"${findProperty('reactNativeReleaseLevel') ?: 'stable'}\""
     }
--- a/app.json
+++ b/app.json
@@ -3,7 +3,7 @@
     "scheme": "private-mind",
     "name": "Private Mind",
     "slug": "private-mind",
-    "version": "1.2.1",
+    "version": "1.3.0",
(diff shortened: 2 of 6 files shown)


NorbertKlockiewicz (approved):

kfaracik:
Superseded by #332, which carries these commits and the rest of the test round. Verified by content before closing: everything on this branch is in release/1.3.0-final.

PR #326, opened 2026-09-11
fix(chat): keep an answer left inside a think block that never closed

kfaracik:
Two commits: the field-report document for 1.3.0, and the first fix from it.

## The defect

A turn succeeded only when `stripThinkBlocks` left something behind. `outsideThinkSegments` returns early on an unterminated `<think>`:

```ts
const close = source.indexOf(THINK_CLOSE, open + THINK_OPEN.length);
if (close === -1) return segments;
```

so everything after the tag is dropped. A thinking model that answers inside its reasoning and never emits the closing tag produced an empty string, and the turn was marked failed.

That cost the answer, not just the label. `persistMessage` lives in the success branch, so the reply had only ever streamed to screen — reported twice from a device, the second time with the disappearance on re-entering the chat called out. The device database agrees: of 35 persisted assistant messages, 16 carry a think block and **none** carries an unterminated one.

## Why the obvious fix is wrong

Accepting any text inside an unterminated block breaks a test this branch already carries:

```
drops the bubble when the model only looped inside an unterminated think block (iPhone SE)
'<think>cząstek cząstek cząstek cząstek cząstek'
```

Both are unterminated blocks holding text. What separates them is whether the text is an answer or a repetition, and the first attempt at this fix could not tell them apart.

## What it does

`loopDetection` already knew the difference and discarded it — `truncateAtRepeatedClause` finds the cut, then salvages the first unit when the cut lands at zero, handing a pure loop back unchanged. `isRepetitionFromTheStart` exposes the detection without the salvage, and `carriesAnswer` accepts an unterminated block only when it does not begin as a loop.

Nothing else moves. The persisted payload was always the raw `finalResponse` including the markup, so a message renders after a reload exactly as it did while it streamed. A closed block with nothing after it still fails the turn.

## Gates

2422 tests across 142 suites, Prettier clean. Both cases are tests and each fails without the change: the answer is kept, the looped one is still dropped.

### DIFF
--- a/__tests__/llmStore.test.ts
+++ b/__tests__/llmStore.test.ts
@@ -1531,6 +1531,26 @@ describe('sendChatMessage', () => {
     expect(content).toContain('</think>');
   });
 
+  it('keeps an answer the model left inside a think block it never closed', async () => {
+    mockInstance.generate.mockResolvedValue(
+      '<think>Weighing the sources. The president of Ukraine is Volodymyr Zelensky.'
+    );
+    useLLMStore.setState({
+      model: baseModel,
+      activeChatId: 1,
+      activeChatMessages: [],
+    });
+
+    await useLLMStore
+      .getState()
+      .sendChatMessage('who is president of Ukraine?', 1, noSources, settings);
+
+    expect(useLLMStore.getState().generationError).toBeNull();
+    expect(useLLMStore.getState().activeChatMessages.at(-1)?.content).toContain(
+      'Volodymyr Zelensky'
+    );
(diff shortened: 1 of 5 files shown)


kfaracik:
Superseded by #332. This branch's doc commit was cherry-picked onto the release branch, so the SHAs differ and merge-base reports it as not contained - the content is there and then some: release/1.3.0-final carries sections 13 through 16 of docs/ISSUES_1.3.0.md, this branch carries 13 through 15. Checked before closing.

PR #327, opened 2026-09-11
fix(chat): settle the pin release and ease the landing after send

kfaracik:
Two of the three chat animation reports in `docs/ISSUES_1.3.0.md`. The third is deliberately untouched — see the last section.

## The blank space was dead code

`blankSpace` is created in `ChatScreen`, reset to 0 in five places and animated to 0 on release. Nothing in the repo ever sets it non-zero, and `KeyboardChatScrollView` treats a zero `blankSpace` identically to its absence. One of the three competing curves in §4 was not competing at all.

## The pin release had one real fight

The floor was removed at **drag start** — a layout change landing whole on one frame, which on Android makes `ReactScrollView` clamp in the same frame and on iOS leaves the offset past the new end — while `onContentSizeChange` fired an animated `scrollTo` with the finger still down.

Now the scroll offset is the only moving quantity:

- a drag arms the release without touching the floor
- on drag or momentum end the list scrolls, on one native curve, to where the natural end will be
- the floor is dropped on the first scroll event that puts it below the viewport, where the layout change cannot be seen
- dragging the other way past that point drops it with no scroll at all

**The property that had to survive:** the floor still holds the question in place while the reply streams in below it. It is never removed during generation.

Supporting changes: the answer row is measured on the inner view so the release target is known before the floor goes; the error view gets the same wrapper (`marginBottom` → `paddingBottom`) so its measured height still carries it; settling skips while the keyboard is visible. The arithmetic is extracted to `components/chat-screen/pinScroll.ts` and `__tests__/pinScroll.test.ts` pins the invariant the release depends on — with the floor, content end equals pin offset plus viewport.

## The landing eases only the last stretch

`scrollToPin` jumps instantly to `MESSAGE_PIN_LANDING_PX` (64) short of the target and animates that stretch alone. `animated: false` over the whole distance was deliberate — the pin must be in place before the reply fills the space below — and this keeps that while giving the eye something to follow. Because the floor fixes the content size, the streaming reply cannot disturb the animated part.

The Android `keyboardDidHide` re-snap lands on the same offset after a send and would have undone the easing; while a pin is active it now routes through the same landing.

## §2 is not addressed, and the diagnosis in the document is wrong

The document blamed the bottom inset being counted twice. That does not survive reading the sources: `react-native-is-edge-to-edge` returns `true` unconditionally on RN 0.83, so keyboard-controller forces `navigationBarTranslucent`, the reported keyboard height already includes the nav-bar region, and `safe-area-context` never reads `ime()`. `theme.insets.bottom` is constant across a keyboard transition, so the baseline re-capture should not fire at all — and a nav-inset-sized mechanism could explain the ±30 px ring at most, not a 135 px jump.

Frames from the original recording show what is actually happening: at frame 107 the conversation content has completed 91% of its move while the keyboard is halfway up and the input bar is still travelling, and the bar only arrives around frame 110. So the content steps while the bar animates — the same family of defect, but the quantity is the bar's height, not the inset.

Left open rather than guessed at. The next measurement is cheap: log `theme.insets.bottom` during a keyboard open, which settles the premise in one line.

## Gates

2433 tests across 142 suites (baseline 2421), `tsc` clean, eslint clean, Prettier clean. Each intermediate commit is green.

Verified by tests: the `pinScroll` arithmetic. Verified by reading only: the release choreography and the landing in `Messages.tsx` — event ordering, native clamp behaviour, and the keyboard-dismiss interaction on Android. Those want a device pass before merge.


kfaracik:
Superseded by #332. Its four commits are ancestors of release/1.3.0-final, checked with merge-base before closing.

Issue #331, opened 2026-09-12
Let the model orchestrate document RAG, image and web search — including repeated web searches in one answer


kfaracik:
Three retrieval sources exist — **document RAG, image, web search** — and the app
picks between them with a fixed rule instead of letting the model decide.

## Today

`RAG_PRIORITY_OVER_WEB_SEARCH` is a constant. Documents win, web search is
skipped, and that is the whole policy. Web search also runs exactly once, before
generation, from a query planned before any answer text exists.

Mutual exclusion is now enforced for all three (documents or image ⇒ no web
search), which is the right stopgap and the wrong long-term design.

## Wanted

The model chooses what to use and in what order, per turn:

- **More than one source in a single answer** — an image and a document, or a
  document plus the web to check whether it is current.
- **Web search more than once per answer.** A first search surfaces a fact that
  raises a second question; the model should be able to search again mid-answer
  rather than answering from one fixed query.
- **Ordering decided by the model, not by a constant.** Read the image first,
  then search for what it shows. Retrieve from the document, then verify a date
  on the web.

## Constraints this has to respect

- One model, on-device, 2048-token context on the smaller profiles. Every extra
  retrieval round competes for the same window.
- Each web round costs a generation for the query plus fetch and extraction time.
- The trace UI currently assumes one search phase per answer.
- Memory: the embedding model, the vision path and the LLM already contend for
  RAM (see the compatibility gate).

## Open questions

- Tool-calling versus a scripted loop with a bounded number of rounds.
- What stops a loop that keeps searching.
- How the trace shows several rounds without becoming noise.

Issue #330, opened 2026-09-11
First launch: the header names a starting model that is not downloaded, and the wait looks like a hang


kfaracik:
On a fresh install the chat header already names a model the device does not
have. Switching to it then looks like an indefinite load.

Reported from a device: "przy pierwszym uruchomieniu przełączenie się na model
Gemma 4 2B ładuje się w nieskończoność."

## Observed on a Pixel 10, clean install of 1.3.2

Straight after first launch, before anything was downloaded:

- The chat header reads **Gemma 4 - 2B**, presented exactly as a selected model.
- `Select a Model` does not offer Gemma 4 - 2B. Scrolling the whole sheet shows
  Qwen 3 - 1.7B, LLaMA 3.2 1B QLoRa / SpinQuant, LLaMA 3.2 3B QLoRa / SpinQuant,
  Qwen 2.5 0.5B / 1.5B / 3B — the selected model is excluded from the list it
  would otherwise appear in.

## Where the name comes from

`constants/default-models.ts` picks a starting model by device RAM:

```ts
const HIGH_END_STARTING_MODELS = [
  'Gemma 4 - 2B',
  'Gemma 4 VL - 2B',
  'Qwen 3 - 1.7B',
];
```

A 12 GB Pixel 10 lands on `Gemma 4 - 2B`, whose catalog entry is
`source: 'remote'` with `modelSize` 2.5 GB on Android and 2.9 GB on iOS. So the
header names a **2.5 GB download that has not happened**, and nothing on screen
says so.

## Why "forever" is the honest description

Nothing in the header distinguishes *downloading 2.5 GB over the network* from
*loading a local file*. A user on a slow connection sees the same spinner for
several minutes as they would see for two seconds, with no progress, no size,
and no cancel. The model-select sheet — the one place that shows a download size
per row — is also the one place that does not list the model being waited on.

## Not established

The indefinite case itself was not reproduced here; what is confirmed is the
state that produces it. Two things would settle it:

1. Whether the download actually completes on a slow link, or whether it stalls.
   A release build strips its console, so this needs a dev build or a network
   capture.
2. Whether `loadModel`'s `await modelOffloadChain` contributes. `loadModel`
   waits on that chain, and `runWithModelOffloaded` — called from
   `VectorStoreContext` — spins in `waitForModelToBecomeIdle` while `isLoading`
   is true. On a *fresh* install that call sits behind `if (downloaded)` and the
[...]

Issue #329, opened 2026-09-11
Chat: the web-search block replays its entrance animation when it hands over to "Searched the web"


kfaracik:
The web-search block re-animates at the exact moment the answer arrives. The
rows it was showing are replaced by a different set, the block changes height,
and the message re-lays out underneath the text the user has just started
reading.

Reported from a device after the 1.3.0 animation work — this one is not fixed by
it and is a separate mechanism.

## What happens at the hand-over

A single `WebSearchBlock` serves two different states, and the switch between
them happens in one commit:

```tsx
const isLiveBlock = isSearching || trace.length > 0;
```

While the search runs, the block is fed the live trace from the store
(`MessageItem.tsx:297-299`). When the turn completes, `resetTrace()` in
`store/webSearchStore.ts:94` clears it:

```ts
resetTrace: () =>
  set({
    ...
    traceExpanded: false,
  }),
```

So in the same update `trace` becomes `[]`, `isLiveBlock` flips to `false`, and
the block's content switches from the live steps ("Searching", "Reading the
pages", "Done") to the persisted source rows.

## Why that re-animates rather than settling

`buildRows` returns an entirely new set of row keys, and the block staggers
every key it has not seen before:

```tsx
const enterDelay = useMemo(() => {
  const delays = new Map<string, number>();
  let fresh = 0;
  for (const row of rows) {
    if (seenKeys.current.has(row.key)) {
      delays.set(row.key, 0);
    } else {
      delays.set(row.key, fresh * STAGGER_MS);
      fresh += 1;
    }
  }
  return delays;
}, [rows]);
```

None of the new keys are in `seenKeys`, so every row is treated as fresh and
enters on the stagger — a full replay of the entrance animation for content the
user was already looking at. The two states are not two renders of one list;
they are two lists that happen to share a component.

The height change compounds it. Several live rows give way to a different number
of source rows, so the block resizes in the same frame, and everything below it
in the message moves. That lands precisely when the answer text starts
streaming, which is the worst possible moment for the list to shift.

## What good would look like

The rows that survive the hand-over should not re-enter, and the ones that go
should leave without dragging the message with them. Concretely, worth weighing:

- **Carry `seenKeys` across the live → persisted switch** so shared rows keep a
  zero delay and only genuinely new rows stagger. Cheapest change, fixes the
  replay but not the resize.
- **Give the block a stable height across the switch**, or animate the height
  the way `WebSearchTraceList` already animates its clip, so the resize is a
[...]

Issue #328, opened 2026-09-11
Web search: a short in-language question finds nothing — "pogoda londyn weekend" on Gemma 4 VL 2B


kfaracik:
A short, ordinary question in the user's own language comes back with nothing.

## Reported

Device report, Gemma 4 VL 2B, web search on:

```
pogoda londyn weekend        =>  no usable results
```

Three content words, a city, a time window. Nothing exotic — this is the shape
of question a phone gets asked most often, and the shape a 2B planner is least
able to rescue.

## What the pipeline does with it

Measured by running the real functions over that exact string.

**The time window makes it a "named day" question.**

```
namesAnotherDay('pogoda londyn weekend')  =>  true
mentionsTime('pogoda londyn weekend')     =>  true
```

`namesAnotherDay` in `utils/calendarFacts.ts` carries the `weekend` stem, so
`getTimeScopeInstruction` in `utils/promptUtils.ts` fires and tells the model:

> Use only the value the page labels with the day the question asks about; a
> figure the page labels "now", "today", "current" or "at the moment" does not
> answer a question about a different day (...) If the page carries no value for
> the day asked about, say so rather than quoting the nearest one.

That instruction is right for "jaka będzie pogoda w sobotę" and wrong here in
one specific way: a weekend is **two** days, and the instruction is written in
the singular. A page carrying Saturday and Sunday rows satisfies the question
but matches no single "the day asked about", and the safest reading of the
instruction is to refuse. Combined with `getMeasurementUnitInstruction`, which
fires on `pogod\w*` and demands a figure printed next to its unit, the model is
given two independent reasons to say it found nothing, before retrieval quality
is even in play.

**An English rewrite of the query is thrown away.**

```
detectQuestionLanguage('pogoda londyn weekend')            =>  pl
detectQuestionLanguage('London weather weekend forecast')  =>  en
sharesLanguageWith('London weather weekend forecast',
                   'pogoda londyn weekend')                =>  false
```

`buildSearchQuery` requires every planned query to share the conversation's
language (`inConversationLanguage`). A plan that fails it costs one extra
generation to replan (`replanInConversationLanguage`) — expensive on a 2B model
on a phone — and if the replan also fails, the queries are **filtered out
entirely**:

```ts
queries: parsed.queries.filter((q) => sharesLanguageWith(q, reference))
```

An empty list then falls through to `verbatim()`, which searches the user's raw
keywords. So the likely path here is: the planner writes a sensible English
[...]

PR #332, opened 2026-09-12
Release 1.3.0: every fix from the test round, in one branch

kfaracik:
Everything that has to land before 1.3.0 ships, in one branch. This supersedes
#321, #326 and #327 — their commits are all in here, so merging this makes those
three redundant rather than conflicting.

Built and tested as 1.3.1 … 1.3.5 on a Pixel 10 and an iPhone 17; the final
commit sets the shipping number back to **1.3.0 / versionCode 69 / CFBundleVersion 5**.

## What is in it

**Answers that were being lost**

- A turn now carries an identity (`generatingMessageLocalId`). `flushStream`,
  the complete phase and the failed phase addressed the assistant message by
  position, so a list rebuilt mid-turn — which the chat route does on focus —
  sent the answer to the previous message and failed the current one.
- An answer the model left inside a `<think>` block it never closed is kept,
  unless the block only holds a repetition.

**Keyboard and send**

- The input lifts from the keyboard's first frame. `keyboardGone` was cleared by
  `keyboardDidShow`, which on iOS arrives after the animation has finished, so
  the bar jumped instead of travelling.
- Scrolling with the keyboard up no longer dismisses it; tapping a message still
  does.
- A sent message travels to the top in one motion, and the conversation no
  longer blanks out for ~200 ms while it does.
- The pin floor follows the keyboard, and is dropped once a finished answer has
  outgrown it.

**Web search**

- The search-model prompt appears on a slow first launch. The wait for the
  embedding status timed out at 8 s and a still-unknown status was read as "no
  prompt needed", leaving the toggle on with nothing behind it.
- That sheet always offers Cancel; it previously had no way out on a device
  capable enough for the model to be required.
- Web search stays off while an image is attached, as it already did for
  documents, with its own notice rather than one about documents.
- Think and Web stay tappable while a model loads — they write chat settings and
  never needed a loaded model.

**Visible version**

- The What's New card reads the installed version instead of a constant, so the
  number on screen is the build in hand.
- Grounding caveat badges are dev-only for now; the caveats are still computed
  and persisted so their frequency can be measured before the decision.

## Known and not fixed

Tracked, deliberately out of scope for 1.3.0: #329 (the web-search block replays
its entrance, and a sent message sometimes stops half-way), #330 (first launch
names a starting model that is not downloaded), #325, #328, #331, and the
`react-native-executorch` abort from §8 of `docs/ISSUES_1.3.0.md`.

## Gates

2448 tests / 144 suites green, typecheck and lint clean, no attribution trailers.
Several of the fixes above are interaction or timing changes that no unit test
can hold — those were verified by hand on the two devices.


kfaracik:
Split into two, as asked: #334 carries the fixes with the version left alone, and #335 on top of it carries the release number and nothing else. Same content as this branch, reviewable in two pieces instead of one.

Issue #338, opened 2026-09-14
Android: an attached image on Gemma 4 VL corrupts the answer and can reset the GPU, rebooting the phone


kfaracik:
On Android, attaching an image to **Gemma 4 VL - 2B** runs the multimodal graph
on the Vulkan backend. That graph misbehaves in two ways on the same build and
the same device: it either returns an answer built from corrupted image
embeddings, or it issues a GPU dispatch that outruns the driver watchdog. When
it outruns, the driver resets the GPU, the display fence never fires, and
surfaceflinger plus system_server go down with it — **the phone reboots.**

Reproduced on `main` at `12ac3a4` (1.3.0 / 69), Pixel 10, clean install.

## Evidence

Dropbox, seven minutes after that build was installed:

```
17:02:43  SubsystemRestart
          subsystem: powervr
          crash reason: powervr - [11957]:[ion.privatemind]:[GUILTY_OVERRUNING]
17:04–17:09  system_server_watchdog ×2, system_app_anr ×8
17:09:25     SYSTEM_RESTART
17:11:28     SYSTEM_BOOT
```

An ANR stack captured during an earlier occurrence, thread `RN_ET_Worker0`:

```
IMG_vkWaitForFences
vkcompute::vkapi::VulkanFence::wait
vkcompute::ComputeGraph::execute
executorch::runtime::Method::execute_instruction
executorch::extension::llm::MultimodalPrefiller::prefillChunk
executorch::extension::llm::MultimodalPrefiller::prefill
executorch::extension::llm::MultimodalRunner::generate_internal
rnexecutorch::models::llm::LLM::generateMultimodal
```

## What isolates it to the image path

Three sends, same model, same session, minutes apart:

| input | result |
| --- | --- |
| image + "What is this?" | `インチ is position is, receivable, denn, is is is is is is` |
| no image, "Name three colours." | `{"response": "Blue, green, and yellow."}` |
| Gemma 4 - 2B (also Vulkan) + web search, long prefill | correct London forecast, 43 s, no GPU event |

So the Vulkan backend as such is fine, and a long prefill is fine. Only the
multimodal graph breaks. The hang is intermittent — the same send produces
garbage on one run and resets the GPU on another.

## Where the defect is

The Android→Vulkan choice originates in `react-native-executorch` 0.9.2:

```ts
modelSource: Platform.OS === 'android' ? GEMMA4_E2B_VULKAN_MM : GEMMA4_E2B_MLX_MM
```

[...]

PR #339, opened 2026-09-14
fix(models): run Gemma 4 VL on the CPU backend on Android

kfaracik:
Closes the crash path in #338 by moving **Gemma 4 VL - 2B** off the Vulkan
backend on Android. On iOS nothing changes — MLX is untouched.

## Why here and not in the library

The defect is upstream: `react-native-executorch` picks the Vulkan export for
`GEMMA4_E2B_MM` on Android, and its Gemma 4 Vulkan support is a port carried on
[their ExecuTorch fork](https://github.com/software-mansion/react-native-executorch/issues/1361)
rather than anything upstream. But the URL our users actually download comes
from **our** catalog, and neither a GPU reset nor a native abort surfaces as a
catchable JS error — so not entering the path is the only lever we have.

Upgrading the library is not that lever either: the Vulkan multimodal `.pte` is
byte-for-byte identical between the `v0.9.0` and `v0.10.0` tags
(`sha256 56c6137e…889b3fd2`), and Gemma 4 multimodal is no longer in the 0.10.2
registry at all.

## What changes

| | before | after |
| --- | --- | --- |
| `modelPath.android` | `e2b/vulkan/gemma_4_e2b_vulkan_8da4w.pte` | `e2b/xnnpack/gemma_4_e2b_xnnpack_8da4w.pte` |
| `modelSize.android` | 4.0 | 4.7 |

Both in `catalog/model-catalog.json` and in the `DEFAULT_MODELS` fallback — a
first launch that cannot reach the catalog would otherwise still be handed the
Vulkan build.

Because the catalog is remote, publishing it to `gh-pages` delivers this to
every 1.3.0 install without a store release. 1.2.1 has no catalog fetch, so
those users can only be reached by updating.

## Cost

- image prefill moves to the CPU and will be slower;
- the URL changes, so anyone who already downloaded the model re-downloads
  4.69 GB, and the old 4.37 GB file is orphaned on disk;
- the RAM gate now reads 4.7 GB, so the model is offered on slightly fewer
  devices.

## Test

`never runs a vision model on the Android Vulkan backend` in
`__tests__/modelCatalogFile.test.ts` — verified red against the previous
catalog (`Array [ "Gemma 4 VL - 2B" ]`).

## Device verification — passed

A release APK built from this branch, with `EXPO_PUBLIC_MODEL_CATALOG_URL`
pointed at this branch's catalog, on a Pixel 10 (the device that reboots on the
Vulkan build). Same photo, same question, same model as the failing run:

> The image provided is a photograph of a workspace or desk setup, likely for a
> designer, developer, or creative professional. The items visible include:
> **Computer Hardware:** Multiple monitors, keyboards, and other computer
> peripherals. **Design/Creative Tools:** Several physical models or objects…
> In summary, it is a depiction of a modern, busy workspace filled with
> technology and design tools.

against the Vulkan build's `インチ is position is, receivable, denn, is is is`.

- no new `SubsystemRestart` in dropbox — the count stayed at the one from the
  Vulkan crash;
- no reboot: uptime ran through the test;
- nothing matching `powervr` or `GUILTY` in logcat;
- first text about 6 s after the send, full answer about two and a half
  minutes. Slower than the GPU path, as expected.

### Two things the run turned up

**Existing owners re-download silently.** The app kept `isDownloaded` from the
old row, and on selecting the model fetched the new file — 3.7 GB at ~15 MB/s —
behind nothing but *"Wait for the model to finish loading."* No progress, no
size, no way to tell it apart from a hang. Worth its own issue.

**Special tokens leak into the answer.** The reply contained
`Office Environment:<unused6226><unused6226>感`. Unrelated to the backend —
these are unfiltered special tokens — but visible to the user. Also its own
issue.


kfaracik:
## ⚠️ Do not publish the catalog yet — the XNNPACK export refuses at 1024 tokens

Continued testing on the Pixel 10, same build, and hit a hard limit on the
second image in one conversation.

### What happened

A chat with one long text answer already in it (~2 800 characters). First image
+ *"Describe this image."* → **answered correctly**, in about eight and a half
minutes. Second image + *"And this one?"* → **"Failed to generate a response."**

From logcat, at the moment of the failure:

```
E/ExecuTorch: Attempted to resize a bounded tensor with a maximum capacity
              of 1024 elements to 1610 elements.
E/ExecuTorch: Error resizing tensor at input 0
E/ExecuTorch: Attempted to resize a bounded tensor with a maximum capacity
              of 1024 elements to 1034 elements.
E/ExecuTorch: Error resizing tensor at input 0
```

Two attempts: the prompt at 1610, then something near 1034 — that second number
looks like our own truncation fallback trimming and still overshooting.

### Why this matters here

`DEFAULT_PROFILE.contextWindowTokens` is **2048**. This export accepts **1024**.
We are budgeting twice what the model will take, so any conversation that grows
past roughly a thousand tokens fails — and an image spends a few hundred of
those at once.

This is very likely the same family as the crash reported on a Galaxy S20 FE at
the **third** attached image, which I had guessed was "an accumulating limit"
without being able to name it. Now it has a name.

### What is not yet established

Whether the **Vulkan** multimodal export has the same 1024 cap. If it does, this
is a pre-existing limit that the backend swap merely made visible, rather than
anything this PR introduces — and on Vulkan the same conversation would have
produced corrupted output or reset the GPU before it got that far. I have not
isolated that, and it decides whether this PR needs to grow or whether the cap
belongs in its own issue.

### Where that leaves the PR

The first image in a conversation works and is correct, which is still strictly
better than an answer built from corrupted embeddings or a phone that reboots.
But it should not go to `gh-pages` until we decide what the context budget for
this model is, because today the app promises 2048 and the file delivers 1024.

### One more measurement worth having

Timing on the CPU path scales badly with history. A fresh chat answered an image
in about two and a half minutes; the same model with one prior answer in the
chat took **eight and a half**. Correct, but slow enough that the "Thinking…"
state needs to say more than it does.

kfaracik:
## Blocker cleared — verified on the conversation that failed

The 1024-token capacity problem I flagged above is fixed in
`fix(chat): keep the prompt inside what the model will actually accept`, and
re-tested on the **same chat on the same device**, with its two images and both
failed turns still in the history.

| step | before | after |
| --- | --- | --- |
| plain text question in that chat | "Failed to generate a response." | answered, ~7 s |
| **third** image + question | would have failed | answered correctly, ~1 min 48 s |
| `maximum capacity` errors in logcat | 4 | **0** |

The third image is the case reported as a crash on a Galaxy S20 FE, and it is
the one that most needed to pass.

## What was actually wrong

Two things, and the second was the larger one.

**The window.** The export takes 1024 tokens; the app budgeted 1536. That is now
declared per model and per platform — iOS runs the MLX build and keeps the
default.

**Every image was re-sent on every turn.** `prepareMessagesForLLM` gave a
`mediaPath` to *any* message carrying an `imagePath`, so a conversation's whole
image history went to the model each time, each image expanding to a few hundred
visual tokens that a character-based estimate cannot see. Three images spend
more context than this model has — which is exactly why the failures arrive on
the second or third image and then never go away, taking plain text questions
down with them.

Only the most recent image is sent now, and the budget charges for it.

## The trade this makes, stated plainly

The model can no longer look back at an earlier picture in the same
conversation — *"what was in the first photo?"* will not work. The text of those
turns is kept, so it still knows they happened and what was said about them.

That is a real loss, and it buys a model that does not stop working after two
pictures. When the context budget is genuinely large enough for more, keeping
several images would be the better behaviour; that belongs in #354 rather than
here.

Answers in a long chat are also noticeably thinner now, because the trimmer is
working against a real 1024-token ceiling rather than an imagined 2048 one. That
is the export's limit showing through honestly instead of as a failure.

## Gates

2459 tests / 145 suites, typecheck clean, lint unchanged at 87 warnings and 0
errors — the same as `main`.

New tests: `__tests__/gemmaVlContextWindow.test.ts` (window matches the export's
capacity on Android, keeps headroom, leaves iOS alone) and three cases in
`promptUtils.test.ts` for the image bound and its charge. The image one was
verified red against the previous behaviour.

kfaracik:
Closing this. The Android Gemma 4 VL workaround here is meant to be solved upstream in react-native-executorch rather than carried as a catalog override on our side — see https://github.com/software-mansion/react-native-executorch/issues/1485 for the current state of the 0.10 upgrade.

Keeping the branch `fix/338-android-vision-on-cpu` rather than deleting it: two commits on it are unrelated to executorch and still stand on their own —

- `a897e68` a catalog size change never reached a model row that already existed
- `a7e0eac` prompts were not clamped to the model's context window

If the upstream fix lands and we still want those, they can be lifted from here.

PR #353, opened 2026-09-15
perf(chat): stop the composer re-rendering on every streamed token

### LINKED ISSUE #352: The composer re-renders on every streamed token, and memoising it is not a one-liner
Found while doing the cheap half of a render audit (#346, #347, #348, #349,
#350). This is the one finding I did **not** act on, recorded so the reasoning
is not lost.

## What happens

`ChatBar` is not memoised, nor is `ChatBarActions`. `ChatScreen` re-renders once
per animation frame while an answer streams — `flushStream` calls `set` at that
rate — so the whole composer subtree re-renders with it: the `TextInput`, the
attachment chips, the Think and Web toggles.

## Why `memo(ChatBar)` alone does nothing

Three of its props get a fresh identity every render:

- `components/chat-screen/useSendChatMessage.ts:94` returns a bare
  `async (...) => {...}`, rebuilt each call;
- `handleThinkingToggle` and `handleWebSearchToggle` in
  `components/chat-screen/useChatScreenActions.ts` are plain function
  declarations.

The remaining props are already stable — `extraContentPadding` is a shared
value, and the rest are `useCallback`ed.

## Why it was left alone

Wrapping the send function in `useCallback` means writing a dependency array for
a ~270-line closure over the model, chat settings, attachments, the web-search
store, the vector store and the router. Getting one entry wrong produces a stale
closure **in the send path** — which is the part of the app the 1.3.0 cycle was
spent stabilising (#227, the first-message interrupt, the pin/floor mechanism).
The cost of a subtle regression there is much higher than the cost of the
re-renders it would save, and the composer is a small tree next to the message
list, which the other PRs already cover.

## If someone picks this up

The safer shape is probably to keep the handlers fresh and hold them in refs,
exposing stable wrappers to `ChatBar`, rather than to chase an exhaustive
dependency list. Whatever the approach, it wants a manual pass on device over
send, prompt suggestions, attachments and the iOS remount key (`composerKey` /
`iosInputKey`) before it lands — `memo` on a component with
`useImperativeHandle` and a remount key is not a change that tests alone will
vouch for.

kfaracik:
>
> **Device smoke test done, 2026-09-15, physical Pixel 10, Gemma 4 VL - 2B.**
> This is the result the section below promised before merging. Every item on
> that list passes:
>
> | checked | result |
> | --- | --- |
> | prompt suggestion fills the composer | the imperative handle still reaches the input through `memo` |
> | Think toggle | fires the correct closure — rejects with *"Thinking cannot be enabled for this model"* |
> | Web toggle | flips and re-renders; the boolean prop crosses the shallow compare |
> | typing | the field updates per keystroke |
> | send | message posts, composer clears and shrinks, send becomes stop |
> | streaming | answer renders normally |
> | stop | halts mid-sentence, the bar returns to idle |
> | attachment sheet | opens (Camera / Photo Library / Document) |
> | switching chat | composer resets, accepts input, no draft leaks across chats |
>
> The `disabled` prop was watched flipping both ways as the model loaded and as
> generation started and stopped, so props do reach the memoised component.
>
> **Two things this does not show.**
>
> The **render-count win is reasoned, not measured.** I tried to record it with
> the React profiler and the CDP session drops as soon as on-device inference
> starts — the profiler and the model share one Hermes runtime. The app itself
> kept streaming throughout; nothing here is a defect in this PR, the
> measurement instrument simply does not survive the workload it would measure.
>
> The **iOS remount key** was verified only in its Android form — the same
> `chatId` change, which resets the composer correctly. A changed `key` remounts
> regardless of `memo`, so the two cannot interact, but no iOS device was used.

---

Closes #352 — which I had opened saying this was too risky to do cheaply. It
isn't, because `hooks/useStableCallback.ts` already exists in this repo and
`ChatSpeechInput`, `useSpeechInput` and `useModelSwitch` already use it. I
missed it when writing that issue.

## What happens

`ChatScreen` re-renders once per animation frame while an answer streams
(`flushStream` sets state at that rate). `ChatBar` was not memoised, so the
whole composer subtree re-rendered with it: the `TextInput`, the attachment
chips, the Think and Web toggles.

## Why `memo` alone would have done nothing

Three props were rebuilt on every render:

| prop | before |
| --- | --- |
| `onSend` | `useSendChatMessage` returns a bare `async (...) => {}` |
| `onThinkingToggle` | plain function declaration |
| `onWebSearchToggle` | plain function declaration |

Every other prop was already stable — a shared value, a ref, `useCallback`
handlers and a `setState`.

## Why `useStableCallback` and not `useCallback`

Writing a dependency array for the send closure means enumerating the model,
chat settings, attachments, the web-search store, the vector store and the
router across ~270 lines. One wrong entry is a stale closure **in the send
path** — the part of the app the 1.3.0 cycle was spent stabilising.
`useStableCallback` holds the latest closure in a ref behind a fixed identity,
so there is no list to get wrong.

## Verification

Tests 2451 passing, typecheck clean, lint unchanged at 87 warnings / 0 errors
(same as `main`).

Device smoke test of send, prompt suggestions, attachments and the iOS remount
key is in progress on a Pixel 10 — `memo` on a component with
`useImperativeHandle` and a remount key is not something the suite can vouch
for. **I will post the result here before this is worth merging.**


kfaracik:
## Device smoke test — passed

Release build of this branch, merged with the other open work, on a Pixel 10
running Gemma 4 VL - 2B. Each path `memo` could have broken:

| path | result |
| --- | --- |
| prompt suggestion fills the input | **pass** — "Can you explain how machine learning works in simple terms?" landed in the field. This is the case that goes through `useImperativeHandle` → `setInput`, the interesting one for `memo`. |
| `Think` toggle | **pass** — tapping it on a model that cannot think showed *"Thinking cannot be enabled for this model."*, so the stabilised handler reaches the latest closure rather than a captured one. |
| send | **pass** — message sent and answered. |
| attachment picker | **pass** — image picked, thumbnail rendered in the composer, sent with a question and answered. |
| composer clears after send | **pass** — field returned to "Ask about anything…". |

No stale-closure symptom in any of them, which is the failure
`useStableCallback` exists to prevent.

PR #355, opened 2026-09-15
feat(web): say why a search is slow while the planner is still running

### LINKED ISSUE #267: Web Search: in-flight "slow device" note when a compute phase exceeds 8 s
## Context

On low/mid-tier SoCs the web-search planner (a full LLM pass via `generateUtility`) is prefill-bound and takes far longer than on flagships — measured on the pilot corpus: **Gemma 4 2B ≈ 16.5 s median on a Snapdragon 865-class phone vs ≈ 3 s on a current flagship**. After the hard 8 s planner cut was removed (only the single-flight guard shipped), the user stares at a hanging **"Defining objectives"** trace row for 15–25 s with no explanation. The bottleneck is the hardware (model size × SoC), not the network or the app — the UI must say so.

**Dependency:** builds on the trace `NoteRow` mechanism (offline/timeout notes), currently in the uncommitted trace work on `feat/web-search` — that must land first.

## Goal

While a web search is in flight and a **compute-bound** phase is still running after **8 s**, show a muted note in the search trace attributing the wait to the model/device combination. This is the **in-flight layer only**.

## Scope (deliberate cuts — MVP)

- In-flight note only. Explicitly **out of scope**: post-search educational messages, persisted latency history / device tiering, warnings when enabling the Web toggle, any adaptive pipeline changes (the planner is not skipped or shortened).

## Behavior

- **Threshold:** 8 s, as a new constant in `constants/web.ts` (do not reuse `WEB_SEARCH_TIMEOUT_MS` — different semantics).
- **Frequency:** fires on **every** qualifying search — no per-session capping or deduplication.
- **Attribution:** must **not** fire when the time is spent in network stages (SERP scrape, page fetches, CAPTCHA). Detection: when the threshold timer fires, the latest trace entry is still `objectives` (planner running) → compute-bound → show the note; otherwise stay silent.
- **Presentation:** muted `NoteRow` in the trace, after the step rows, visible **only while the search is in flight**; dropped from the finished trace regardless of outcome (results / timeout / no results).
- **Copy (EN):** `A large model takes longer on this phone` — attribution always names the model-on-this-device combination, never the phone alone.

## Acceptance criteria

- [ ] With Web enabled and the planner still running 8 s after the search starts, a muted note appears in the trace under the current step.
- [ ] The note appears on every qualifying search (no suppression across searches).
- [ ] The note does not appear when the search is already past planning at the 8 s mark (network-bound waits never trigger it).
- [ ] The note disappears once the search finishes; it is never rendered in a finished/collapsed trace.
- [ ] Offline and timeout notes keep their current behavior and precedence.
- [ ] The threshold timer is cleared on search completion and on abort/interrupt (no stray timers).
- [ ] Unit tests: `buildRows` shows the note during a live search and drops it when finished; the trigger fires only when the last trace entry is `objectives` at the threshold.

## Implementation sketch (reference)

- `constants/web.ts` — threshold constant (e.g. `WEB_SLOW_DEVICE_NOTE_MS = 8000`).
- `utils/web/runWebSearch.ts` — extend the `WebSearchProgressEvent` union with `'slow'` (pushed from outside the pipeline, same pattern as `timeout`).
- `store/llmStore.ts` (`sendChatMessage`) — a timer next to the 90 s `searchTimeout`; on fire, read `useWebSearchStore` and push `{ type: 'slow' }` if the search is active and the last trace entry is `objectives`; clear it in the same `finally`.
- `components/chat-screen/webSearchTrace.ts` — map `'slow'` to a muted `NoteRow` in the live branch only. `WebSearchTraceRow` already renders muted notes — no changes needed.

This exact scope was prototyped end-to-end and reverted on request; no code remains on the branch — the sketch above reflects the verified approach.

kfaracik:
>
> **Verified on a physical Pixel 10, 2026-09-15**, Gemma 4 VL - 2B, Web on, two
> consecutive runs in separate chats. This supersedes the *"not verified on
> device"* note below, which was true when the PR was opened.
>
> Both acceptance criteria that needed a device hold. While the planner was
> still running:
>
> ```
> Button     "Checking whether to search…"  id="web-search-block"
> StaticText "A large model takes longer on this phone"
> ```
>
> and once the block handed over to *Searched the web*, the note was gone from
> the trace with no other row disturbed.
>
> To see it on a flagship I lowered `WEB_SLOW_DEVICE_NOTE_MS` to 1200 on a
> throwaway branch. **That tweak is not in this PR** — it keeps 8000.
>
> Still unexercised: the network-bound case, where the note must stay silent.
> Both runs were planner-bound at the threshold, which is the branch the note
> exists to fire on.

---

Closes #267.

## Why

On a mid-tier SoC the web-search planner is prefill-bound: Gemma 4 2B takes about
16.5 s on a Snapdragon 865-class phone against about 3 s on a current flagship.
After the hard planner cut was removed, the trace sits on **"Deciding what to
search for"** for 15-25 s with nothing to explain it, and the wait reads as a hang.

The bottleneck is the model on this device, not the network — so the UI should
say that, and only when it is true.

## What it does

At 8 s (`WEB_SLOW_DEVICE_NOTE_MS`) a timer checks whether the newest trace entry
is still `objectives`. If it is, the planner is still running and a muted note
appears under the current step:

> A large model takes longer on this phone

If the search is already past planning, the wait belongs to SERP scraping, page
fetches or a CAPTCHA, and the note stays silent — attribution is the whole point
of the feature, so a network-bound wait must not be blamed on the phone.

The note describes what is happening *now*, not what happened, so it lives only
while the search is in flight and never reaches a finished or reopened trace.
Offline and timeout keep their precedence over it.

## Shape

- `constants/web.ts` — `WEB_SLOW_DEVICE_NOTE_MS`, its own constant rather than a
  reuse of `WEB_SEARCH_TIMEOUT_MS`; different semantics, and they will drift.
- `utils/web/runWebSearch.ts` — `'slow'` joins the progress-event union. It is
  pushed from outside the pipeline, the same way `timeout` is.
- `components/chat-screen/webSearchTrace.ts` — `isStalledOnPlanning` names the
  trigger rule, and `buildRows` maps `slow` to a muted `NoteRow` in the live
  branches only. `WebSearchTraceRow` already renders muted notes; no change there.
- `components/chat-screen/useSendChatMessage.ts` — the timer, cleared in the same
  `finally` that clears the searching flag, so an abort or an interrupt takes it
  down too.

## Tests

Five cases in `webSearchTrace.test.ts`. Each of the three guards was verified by
mutation — removing `isSearching`, `!timedOut` or `has('slow')` individually turns
a different test red:

| guard removed | test that fails |
| --- | --- |
| `has('slow')` | shows under the current step while the planner is still running |
| `isSearching` | is gone from the finished trace whatever the outcome |
| `!timedOut` | yields to the notes that explain a failure |

2456 passing, `tsc` / eslint / prettier clean.

## Not covered

The timer wiring itself. The send hook's test harness has no web-search path, and
building one to assert two lines guarded by `try/finally` would mean rebuilding a
mock shared with every other test in that file. Called out here rather than left
for a reviewer to discover.

Not verified on device either: on a Pixel 10 the planner finishes in about 3 s, so
reproducing the note needs the threshold lowered and a build. Worth one round if
someone is already testing on a mid-tier phone.

## Ordering

Touches the same file as #345 and conflicts with it textually in
`webSearchTrace.test.ts` (both append at the end; the production file merges
cleanly). Land #345 first and this rebases without trouble.


kfaracik:
Verified on a physical Pixel 10, Gemma 4 VL - 2B, Web on.

The PR said this was unverified on device because the planner finishes in about
3 s on a flagship. For the round I lowered `WEB_SLOW_DEVICE_NOTE_MS` to 1200 on a
throwaway integration branch — **that tweak is not in this PR**, which keeps 8000.

While the planner was still running, the note appeared under the current step:

```
Button     "Checking whether to search…"  id="web-search-block"
StaticText "A large model takes longer on this phone"
StaticText "Thinking…"
```

Once the search finished and the block handed over to *Searched the web*, the
note was gone from the trace, with no other row disturbed:

```
Button     "Searched the web"  id="web-search-block"
StaticText "Thinking…"
```

Reproduced on two consecutive runs in separate chats. Both acceptance criteria
that needed a device — appears under the current step while planning, absent from
the finished trace — hold.

Still unexercised on device: the network-bound case, where the note must stay
silent. Both runs were planner-bound at the threshold, which is the branch the
note is meant to fire on.

Issue #325, opened 2026-09-11
web search: replace per-language keyword lists with signals that scale to every language


kfaracik:
## Summary

Web search decides what a question *is* — a price question, a date question, a request for a procedure, a refusal to search — by matching hand-written keyword lists per language. Every new language means editing every list, and no one ever does, so in practice the lists encode **English and Polish** and everyone else gets the fallback path.

Polish is **3.6 %** of users. English is 41 %, Hindi 12.8 %, Urdu 5.3 %, German 4.9 %. The two languages the code knows best are the first and the sixth.

This issue is to replace keyword matching with signals that do not need a list per language, where the measurement says that is safe — and to say explicitly where it is **not** safe, because one of these lists must stay.

## Where it lives today

| site | what it matches | languages covered |
| --- | --- | --- |
| `utils/queryTerms.ts` | stopwords, 14 sets (`EN`, `PL`, `HI`, `UR`, `DE`, `PT`, `ES`, `FR`, `RU`, `AR`, `FA`, `ID`, `TR`, `IT`) | 14 |
| `utils/questionLanguage.ts` | `WORDS` + `LETTERS` — language detection itself | 26 / 15 |
| `utils/calendarFacts.ts:31-35` | date and weekday words | ~4 |
| `utils/messageSources.ts:588` | "when" — `kiedy\|when\|wann\|quand\|cuándo\|quando\|когда\|कब` | 8 |
| `utils/messageSources.ts:590` | "how much" — `ile\|ilu\|kurs\|cen[ay]\|koszt\|how much\|price\|cost\|rate\|population\|сколько\|कितन` | 7 |
| `utils/web/figureGrounding.ts:18` | price words — `price\|cena\|cenie\|worth\|costs?\|kosztuj\|wart\|sprzedaj` | **2** |
| `utils/web/figureGrounding.ts:152` | trend/change words | **2** |
| `utils/promptUtils.ts` `PROCEDURE_MARKERS` | recipe / steps / list | **2** |
| `utils/promptUtils.ts:434` | "what's worth doing" | **2** |
| `utils/web/buildSearchQuery.ts:529` | possession pattern `jaki\w*\|jaka\|ile\|co\|który + ma\|mają\|posiada` | **1 (pl)** |

The asymmetry is the point. The stopword list is maintained for 14 languages; the detectors that decide **prompt behaviour** are maintained for two. A Hindi user gets their stopwords stripped and then hits a price detector that has never heard of Hindi.

[...]

PR #356, opened 2026-09-15
test(i18n): fail when a new per-language word list appears

kfaracik:
>
> **Test-only change; no production code is touched, so there is no device
> behaviour to verify.** CI is the whole verification surface here and it is
> green: 2453 passing across 145 suites, `tsc` / eslint / prettier clean.
>
> The guard is self-demonstrating — it runs against the current tree and passes,
> and the two `KNOWN_SITES` entries record the verdict for each existing list,
> so a new one fails loudly rather than being waved through.
>
> Zero production risk: nothing ships to a user from this PR.

---

Delivers the last box of #325's scope — *„Add a lint or a test that fails on a
new per-language word list, so this does not grow back"*.

## What it does

A test walks `utils`, `constants`, `components`, `store` and `hooks` for
constants tagged with a language code in either shape — a prefix (`PL_STOPWORDS`)
or a suffix (`NO_ANSWER_PATTERNS_EN`) — and compares the result against a declared
inventory. A new list fails wherever it lands: in a file already on the list, or
in a file that was not. Both directions are mutation-verified.

## Each site carries a verdict, not just a name

The two sites in the tree today have **opposite** verdicts, and a bare inventory
would hide that:

| site | verdict |
| --- | --- |
| `utils/queryTerms.ts` — 14 stopword sets | **kept on purpose.** They clean a query and decide no behaviour. |
| `constants/citations.ts` — 8 constants | **debt.** Refusal detection is English and Polish only. |

## Two findings that came out of writing it

**1. The stopword union from #325's first scope box is not safe.** Measured on
the 103-question corpus it changes one answer, exactly as the issue says — but the
corpus contains no case where a word is content in one language and a function
word in another. Outside it, the union eats `war`, `hat` and `die` (German) and
`son` (Spanish) out of English questions. Detail in
[#325](../issues/325#issuecomment-5682237054).

**2. `constants/citations.ts` is not in #325's table at all, and it is worse than
the stopwords.** `looksLikeNoAnswer` recognises a refusal in **2 of 15 languages**:

```
RECOGNISED (2/15): en, pl
MISSED (13/15): de, es, fr, pt, it, ru, tr, id, hi, ur, ar, fa, zh
```

The consequence is user-visible. Same refusal, same two sources, through the real
`pickCitationsByAnswer`:

```
en: used=[]
de: used=["Wetter London","Met Office"]
hi: used=["Wetter London","Met Office"]
```

An answer that said it had no information still gets its sources cited underneath,
for everyone outside English and Polish — roughly 45 % of installs. Filed
separately; **not** fixed here, because the obvious language-neutral signal
(zero term overlap between answer and sources) is already used with the opposite
sense at `messageSources.ts:424` to protect answers written in a different
language from their sources. That is a design question, not a patch.

## Limit, stated plainly

The scan reads declarations, so an alternation written inline inside a regex —
`kiedy|when|wann|quand` — still slips past. That is the larger half of what #325
catalogues and it needs a different instrument. This closes the shape that keeps
reappearing.

`LFM_2_5_EMBEDDING_MODEL_ID` is excluded by name: it ends in the Indonesian code
and means identifier.

## Gates

2453 passing across 145 suites, `tsc` / eslint / prettier clean. Test-only change,
no production code touched.

Refs #325


kfaracik:
Reviewing my own PRs critically, this is the one I am least sure earns its place, and I would rather say so than let it sit.

**What it actually catches is narrow.** It matches declarations shaped `<LANG>_NAME` or `NAME_<LANG>`. The larger half of #325 — an alternation written inline inside a regex, `kiedy|when|wann|quand` — still walks straight past, as the PR body admits. So the guard covers the shape that is easy to detect rather than the shape that is common.

**It carries maintenance.** `KNOWN_SITES` has to be kept in step with the two existing sites, and `NOT_A_LANGUAGE` already needs one exception (`LFM_2_5_EMBEDDING_MODEL_ID`, which ends in the Indonesian code). Every future legitimate constant that ends in a two-letter code will need another.

**Its real value was the investigation, not the guard.** Writing it is what surfaced that `looksLikeNoAnswer` covers 2 of 15 languages, which became #357 and is a genuine user-visible bug. That finding is now recorded independently of this file.

So the honest question for a reviewer: is 129 lines of scanner worth catching one naming shape, given the codebase deliberately keeps 14 per-language stopword arrays in `queryTerms.ts` and the guard has to allowlist them anyway?

I would be fine with any of: merging as is, shrinking it to just the `KNOWN_SITES` verdicts as documentation, or closing it and keeping #357. I do not think it should merge simply because it is green.

Issue #268, opened 2026-07-29
Render tappable source badges instead of plain-text "Source X" mentions in answers


kfaracik:
## Problem

Model replies sometimes reference sources verbatim, e.g. *"According to Source 2, …"*. The user sees an opaque label instead of an identifiable, tappable reference. Expected: an inline badge/chip showing the source — document name with an icon for RAG documents, favicon + title + link for web pages.

## Why it happens

- The prompt explicitly labels context blocks as `Source N`: `utils/contextUtils.ts` (`sourceBlock`: `--- Source ${index + 1}: ${name} ---`), and the system prompt describes them to the model as `"Source N: <name>"` (`utils/promptUtils.ts`, `getContextInstruction`). Small models naturally echo these labels into the visible answer.
- The existing citation pipeline only handles `[n]` markers (`CITATION_PATTERN` in `constants/citations.ts`) and only **strips** them from the rendered text (`utils/citations.ts`, applied in `components/chat-screen/MessageItem.tsx`). Plain-text "Source N" phrases are neither detected nor replaced with anything.

Note: the `Source N` prompt labeling already exists on `main` (RAG documents), so this is not specific to `feat/web-search` — that branch only extends it to web pages.

## Proposed scope

1. Inline badge/chip rendered in the answer: for a document — file name + icon; for a web page — favicon + title, tappable (opens the link or the sources sheet). Reusable building blocks already exist: `components/chat-screen/WebFavicon.tsx`, `components/chat-screen/SourcesSheet.tsx` and the `onShowSources` handler.
2. Detect and map both `[n]` markers and "Source N" phrases to the concrete `SourceDocument` (available per message via `sourceDocuments`).
3. Optionally: strengthen the prompt so the model cites with markers instead of prose — not sufficient on its own with small models, so post-processing is needed regardless.

## Considerations

- Rendering badges requires hooking into the Markdown pipeline (`components/chat-screen/MarkdownComponent.tsx`) — decide whether the badge is inline in the text flow or listed under the paragraph.
- Should land after `feat/web-search` is merged, since web sources and favicon infrastructure live there.

PR #358, opened 2026-09-15
fix(chat): cite the source the model named, not the one it reads like

kfaracik:
>
> **Verified on a physical Pixel 10, 2026-09-15 — and the device round found a
> defect in this branch, fixed in `c272835`.**
>
> The ordinal was written to the database but never read back:
> `parseSourceDocuments` rebuilds each source field by field and the new field
> was missing from that mapping. With `indexByOrdinal` empty, `namedSourceIn`
> returned `null` for every block and the whole citation path fell back to stem
> overlap — the behaviour this PR exists to replace. It worked only while a
> message was still in memory and died the moment the chat was reopened.
>
> A live search cannot catch that; both runs read from the in-memory message.
> It showed up only by seeding a message whose Polish wording points at source 1
> while its marker names source 2, then reopening the chat.
>
> After the fix, the same seeded message reloaded from the same rows renders
> three blocks and three badges: the marked block cites the source it named, the
> unmarked one falls back to overlap, and the `Source N` prose form resolves too.
> `[n]` is stripped from the visible text. Real web searches were also checked —
> five sources, ordinals 1–5 persisted in prompt order, read straight back out
> of the on-device database.
>
> `2a45470` then makes the guard structural rather than ordinal-specific: a
> `Required<SourceDocument>` fixture means the next field added to the type
> breaks the build until it is both listed and read back. Both halves were
> verified by adding a throwaway field and watching each stop the build in turn.
>
> 2464 tests green, `tsc` clean, CI green on the head commit.
>
> One observation, not a blocker: `Source 2` written as prose stays visible in
> the rendered text — `stripCitations` only removes the bracketed form. It is
> read correctly for attribution, just not cleaned up.

---

Groundwork for #268, and a correctness fix that falls out of it.

## What I found first

The badge machinery #268 asks for **already exists**. `attributeSourcesByBlock`
splits an answer into blocks and attributes a source to each;
`AttributedAnswer` renders a `DominantSourceBadge` under every attributed block.
What was missing is not the UI — it is that **the app throws the model's own
citation away and then guesses**.

Order of operations on `main`:

1. `MessageItem` runs `stripCitations`, deleting every `[n]`.
2. `AttributedAnswer` hands the stripped text to `attributeSourcesByBlock`.
3. Attribution re-derives the source from stem overlap, needing
   `CITATION_MIN_MATCH_SCORE = 2` hits to commit.

So a reply that said *exactly* which source it used has that answer discarded,
and a short sentence can then land on a different source, or on none.

## Why it could not simply be read back

`SourceDocument` never recorded the number the model was shown. Deriving it at
render time does not work: `useMessageSources` dedupes, and attribution filters
to sources still marked `used`, so by the time the badge is drawn the array no
longer lines up with the prompt. Hence two commits.

**`feat(rag): save the number each source was shown under`** records the ordinal
where the block is written — documents in `contextUtils`, web results in
`webResultsToContext`, the currency-quote path in `runWebSearch`. Results that
never reached the prompt get none, because the model was never shown a number
for them. The field is optional, so existing messages are unaffected and there
is no migration.

The test is the invariant rather than a restatement: for every source carrying
an ordinal, the context must contain `--- Source ${ordinal}: ${name} ---`.
Shifting either builder by one turns it red.

**`fix(chat): cite the source the model named, not the one it reads like`** makes
a named source beat the overlap vote. Both shapes count — the bracketed marker
the prompt asks for, and the prose label small models write instead (*"According
to Source 2"*), which #268 reports they do. When a block names several, the one
it repeats most wins. Stripping moved into the attribution pass, the only place
that sees both the raw text and the blocks, so markers reach the matcher and
still never reach the screen.

The fallback is untouched and still carries every case the marker does not: no
marker, a number naming a source that was dropped, an ordinal missing because
the message predates this.

## Verified by mutation

| mutation | test that goes red |
| --- | --- |
| named source no longer beats the vote | follows a bracketed marker away from the better-overlapping source; follows the prose label |
| web ordinal off by one | matches the number printed in that source's own header |
| document ordinal off by one | saves each document under the number its own header printed |

An existing test also caught a real regression while I was writing this — moving
the strip into `AttributedAnswer` had keyed it on `webResults`, which is empty
for a document-only message, so `[1]` survived into the rendered text. That is
why `hasSources` is now passed down explicitly.

## Deliberately not done

**The prose label stays visible.** Deleting *"Source 2"* mid-sentence leaves
"According to , the forecast is 14C", and replacing it with a name the block was
only *guessed* to belong to would print a claim the app cannot stand behind.
Turning it into the tappable chip #268 asks for is the follow-up this unblocks —
the ordinal now makes that resolution exact rather than a guess.

**Not verified on device.** Every change here is pure logic with unit coverage,
but the visible effect — the right badge under the right paragraph when the model
cites explicitly — has not been watched on a phone.

## Gates

2461 passing across 144 suites, `tsc` / eslint / prettier clean. Each commit was
checked out and run on its own; both are green independently.

Refs #268


kfaracik:
The ordinal half is verified on a physical Pixel 10, Gemma 4 VL - 2B, Web on.

A real web search wrote five sources, and every one carries the number it was
shown under, in prompt order, read straight back out of the on-device database:

```
ordinal=1  read=True   used=True   Londyn Pogoda na Weekend - Pogoda w INTERIA.PL
ordinal=2  read=True   used=False  Pogoda Londyn - dziś, jutro, 7 dni i prognoza…
ordinal=3  read=False  used=False  Prognoza pogody na weekend dla lokalizacji Lo…
ordinal=4  read=False  used=False  London Weather This Weekend (Fri-Sun)
ordinal=5  read=False  used=False  Pogoda Londyn - Prognoza pogody godzinowa na…
```

Note ordinals 3–5: results that were surfaced but never fetched still carry a
number, because they were still written into the prompt as `Source N` blocks.
That matches what the code does — the field is omitted only for results that
never reached the prompt at all, which this search did not produce.

**The attribution half is not verified.** It only shows itself when the model
writes `[n]` or *"Source N"*, and Gemma 4 VL - 2B did neither across two runs —
one collapsed to `text="jest"`, the other wrote a long English paraphrase with no
citation marker of any kind. The fallback carried both, which is the intended
behaviour, but it means the new branch went unexercised. Seeing it needs a model
that cites, or a seeded message.

Detail on what those two runs revealed about the sources themselves is in
[#328](../issues/328#issuecomment-5684691990).

kfaracik:
## The attribution half is now verified — and it was broken

My earlier comment said the ordinal half held on device but the attribution
branch went unexercised, because Gemma 4 VL - 2B never wrote a citation marker.
I seeded one instead. It found a real defect in this branch, fixed in c272835.

## The fixture

Two web sources, and an answer whose wording points one way while its marker
points the other:

| | |
|---|---|
| Source 1 | *Londyn — pogoda na weekend (INTERIA)*, Polish passage: `Sobota 19.09 temperatura 21°C, opady przelotne` |
| Source 2 | *Met Office — London weekend forecast*, English passage: `Saturday 19 September highs of 21C with scattered showers` |

```
W sobotę w Londynie temperatura sięgnie 21°C, a opady będą przelotne [2].

W niedzielę zapowiada się pogodnie.

Source 2 podaje, że w niedzielę wystąpią rozpogodzenia.
```

Every content word in the first block overlaps source 1; only the `[2]` names
source 2. Stem overlap alone picks source 1, this branch should pick source 2 —
so the fixture separates the old behaviour from the new one. Confirmed against
`attributeSourcesByBlock` before it went anywhere near the phone.

## What the phone showed

Seeded into the chat database, reopened on a Pixel 10, the message rendered as
**one block with one badge — `Source: Londyn — pogoda na weekend (INTERIA)`**.
Source 1. The overlap answer. Exactly what this branch is meant to stop.

## Why

`parseSourceDocuments` rebuilds every source field by field, and `ordinal` was
never added to that mapping. It is written on save (`def2b44`) and dropped on
load.

`attributeSourcesByBlock` builds `indexByOrdinal` from that one field. With it
empty, `namedSourceIn` returns `null` for every block and the whole path falls
back to stem overlap. So the fix worked only while the message was still in
memory and died the moment the chat was reopened — which is the state a user
spends nearly all their time in. Two runs of a live search would never have
caught it; both read from the in-memory message.

## After the fix

Same seeded message, same phone, reloaded from the same rows:

```
StaticText "W sobotę w Londynie temperatura sięgnie 21°C, a opady będą przelotne."
Button     "Source: Met Office — London weekend forecast"   id="dominant-source-badge"
StaticText "W niedzielę zapowiada się pogodnie."
Button     "Source: Londyn — pogoda na weekend (INTERIA)"   id="dominant-source-badge"
StaticText "Source 2 podaje, że w niedzielę wystąpią rozpogodzenia."
Button     "Source: Met Office — London weekend forecast"   id="dominant-source-badge"
```

Three blocks, three badges. The marked block cites the source it named, the
unmarked one still falls back to overlap and lands on source 1, and the
`Source N` prose form resolves too. `[2]` is stripped from the visible text.

## Guard

Two tests in `chatRepository.test.ts`, in the `source provenance` block that
already exists to catch fields that do not survive a reload — `ordinal` was
added without joining it. One asserts the round trip for web and document
sources alike (document chunks are numbered too, so the field is not web-only);
the other asserts the consequence, that a reloaded answer still cites the source
it named. Both fail with the one-line mapping removed, and no existing test
changes either way.

## One observation, not a blocker

`Source 2` written as prose stays visible in the rendered text — `stripCitations`
only removes the bracketed form. It is read correctly for attribution, it just
is not cleaned up. Whether that is worth handling depends on how often a model
writes the label instead of a marker; I have not seen Gemma do either.

kfaracik:
Follow-up: the guard in `2a45470` is now structural rather than ordinal-specific.

Naming `ordinal` only protects the field that was already lost. The trap is that the reader lists fields one by one while the writer serialises the whole object, so the *next* field added to `SourceDocument` disappears the same way.

A `Required<SourceDocument>` fixture makes the type the checklist: add a field and `tsc` fails until the fixture names it, then the round-trip assertion fails until `parseSourceDocuments` reads it. I verified both halves by adding a throwaway field and watching each stop the build in turn.

2464 tests green, tsc clean.

PR #366, opened 2026-09-18
chore(deps): react-native-executorch 0.10.2 via the legacy API, native libs pinned to v0.10.4-libs

### LINKED ISSUE #344: Upgrade react-native-executorch 0.9.2 → 0.10.x: blocked on a three-way dependency cascade
Scoping the upgrade asked for while investigating #338. It is not a version
bump — 0.10.0 is [a ground-up rewrite of the library](https://github.com/software-mansion/react-native-executorch/releases/tag/v0.10.0),
and the version constraints around it collide in three places at once.

## The blockers

**1. worklets: reanimated and executorch want different majors**

| package | requires `react-native-worklets` |
| --- | --- |
| `react-native-reanimated@4.3.1` (ours) | `0.8.x` |
| `react-native-executorch@0.10.2` | `>=0.10.0 <0.13.0` |

We ship worklets `0.8.3`. There is no version satisfying both, so RNE 0.10
forces a reanimated upgrade — into the animation stack the 1.3.0 keyboard and
scroll work sits on.

A consistent set does exist on our RN 0.83.4:

| | react-native | worklets |
| --- | --- | --- |
| `reanimated@4.5.5` | 0.83 – 0.86 | 0.10.x – 0.11.x |
| `reanimated@4.6.0` | 0.83 – 0.87 | 0.12.x |

Both satisfy RNE 0.10.2. Expo SDK 55 pins reanimated, so `expo-doctor` will
have an opinion.

**2. `@react-native-rag/executorch` has no 0.10 release**

Latest is `0.9.0`, peer-requiring `react-native-executorch: ^0.9.0`. We use its
`ExecuTorchEmbeddings` in `utils/lfmEmbeddings.ts` — the embedding model behind
document RAG and web-search ranking. Either that package ships 0.10 support, or
we replace the wrapper with the new RNE text-embeddings API ourselves.

**3. The import surface moved**

We import `initExecutorch`, `LLMModule`, `SpeechToTextModule`, `ResourceFetcher`,
`WHISPER_TINY_EN` and the model constants from the package root. In 0.10 these
live under a `./legacy` subpath, with the new API being `use<Task>` hooks and
`create<Task>` pipelines over a `models` registry. Legacy imports are the cheap
route, the rewrite is the real one.

## It would probably not fix #338 anyway

The Vulkan multimodal `.pte` is byte-for-byte identical between the `v0.9.0`
and `v0.10.0` tags (`sha256 56c6137e…889b3fd2`), and Gemma 4 multimodal is no
longer in the 0.10.2 registry at all. What does change is the runtime —
0.10.x libs are built against ExecuTorch 1.4.1 against 1.3.1 for 0.9.x — so the
same file on a new runtime is a real experiment, just not a cheap one.

#339 takes the image path off Vulkan without any of this.

## Suggested sequence

1. reanimated 4.3.1 → 4.5.5 with worklets 0.11.x, on its own, regression-tested
   against the keyboard and scroll behaviour in #216, #307 and 1.3.0's send
   animation. This is the risky step and deserves its own PR.
2. Resolve the RAG embeddings question — upstream release or our own wrapper.
3. RNE → 0.10.2 with `./legacy` imports, then a device smoke test of the same
   Vulkan `.pte` to see whether ExecuTorch 1.4.1 changes the #338 behaviour.
4. Migrate to the new API surface, if it earns its keep.

### LINKED ISSUE #1475: Gemma 4 multimodal on Vulkan starves the display for minutes on PowerVR (Pixel 10)
### Description

One image through `gemma4-e2b-multimodal` — the **Vulkan** export, which
`GEMMA4_E2B_MM.modelSource` selects on Android — held the GPU for **12 minutes
without emitting a token** on a Pixel 10 (PowerVR D-Series DXT-48-1536, driver
`25.3@6908880`), with the device unable to present a frame the whole time. I
force-stopped it; it had not finished. Export tag and runtime match, and the
file was deleted and re-downloaded before the run.

| export | input | result |
| --- | --- | --- |
| gemma4-e2b-mm **vulkan** | text | `{"answer": "Blue, green, and red."}` — seconds |
| gemma4-e2b-mm **vulkan** | one 768×768 image | **no output in 12 min, display frozen** |
| lfm2.5-vl-450m **xnnpack** | same image | "A circle is in the image. The color is red." — seconds |

### The display is starved; the GPU is not hung

GPU held a steady ~1.29 W throughout. The app's clock froze at 12:27 while the
system clock read 12:39; `Choreographer: Skipped ~480 frames!` every four
seconds; `Fence: waitForever … drv(pvrsrvkm) status(0)`; and `screencap` started
at 12:31 returned only at 12:37, with a frame still showing 12:27.
Force-stopping the app unfroze the display at once. The same model answers an
image on its XNNPACK export in ~1 min 48 s.

### One OS build earlier, this rebooted the phone

On Android 16, dropbox and the kernel log:

```
crash reason: powervr - [11957]:[ion.privatemind]:[GUILTY_OVERRUNING]
pvrsrvkm 34f00000.gpu0: GPU reset reason=GUILTY_OVERRUNING pid=5037
```

48 `Compute dispatch` commands queued, the display fence never fired, then
`system_server_watchdog` ×2 and `SYSTEM_RESTART`. The ANR trace on thread
`RN_ET_Worker0` sat in `IMG_vkWaitForFences` under:

```
vkcompute::ComputeGraph::execute
executorch::extension::llm::MultimodalPrefiller::prefill
rnexecutorch::models::llm::LLM::generateMultimodal
```

That build also returned corrupted output when it returned at all:
`インチ is position is, receivable, denn, is is is is is`

### A lead: your bundled ExecuTorch predates the PowerVR fix

pytorch/executorch#17299 — same GPU, same device class — was root-caused to
`ComputeGraph::prepack()` batching prepack dispatches into one command buffer,
where on PowerVR only the first prepacked constant gets correct data. Fixed by
pytorch/executorch#22147 (2026-08-26, tested on a Pixel 10), which deleted
`glsl/no_op.glsl` and added `Context::register_barrier`.

Both your snapshots predate it — 0.9.2's `libexecutorch.so` and
`v0.10.0-libs`' `libvulkan_executorch_backend.so`, which `0.10.2` pins via
`nativeLibsVersion`. In each, `strings | grep -c no_op` is **18** and
`nm -DC | grep -c register_barrier` is **0**, against 18 other `barrier`
symbols, so the table is not stripped; `no_op.glsl` is gone from ExecuTorch
`main`. **Upgrading to 0.10.2 does not pick the fix up.**

Not a claim of causation — and one fact cuts against it: this driver is the one
on which #17299's reporter stopped reproducing, though their models were 5–6 MB,
under `prepack_threshold_nbytes`, so `prepack()` never split the buffer. This
one is 4 GB.

### What would help

1. Run a VLM image prompt on any PowerVR device against current `main`.
2. Rebuild the Vulkan artifacts against an ExecuTorch containing #22147 — the
   only known PowerVR fix absent from your binaries.
3. Look separately at GPU allocation lifetime (see performance logs).
4. Until then, consider not defaulting Android to a Vulkan multimodal export.

### Not established

Whether the shipped `LFM2_5_VL_*.VULKAN_8DA4W` exports behave the same: they
exist only from `v0.10.0`, so a failure on the 0.9.2 runtime would not separate
PowerVR from a backend too old for a newer export — though a `v0.10.0` XNNPACK
export does load and answer correctly there. Whether the prefill ever finishes,
or always ends in a reset, is also unknown, and nothing is logged: the stall is
silent.

### Steps to reproduce

1. On a PowerVR device (Pixel 10 / Tensor G5 here), load `GEMMA4_E2B_MM` — on
   Android it resolves to the Vulkan export
   `gemma-4-multimodal/resolve/v0.9.0/e2b/vulkan/gemma_4_e2b_vulkan_8da4w.pte`
2. Wait for the model to finish loading.
3. Send a text-only prompt ("Name three colours."). It answers in seconds, so
   the model and the Vulkan backend work.
4. Send one image (a 768×768 PNG of a red circle on white sufficed) with
   "What shape and colour is in this image?"
5. Watch the device, not the app: the UI stops updating within seconds,
   `screencap` blocks for minutes and returns a stale frame, and `dumpsys gpu`
   shows the app holding ~3.6 of 4.0 GB of GPU.
6. No token appears. After 12 minutes `am force-stop` recovers the display.

### Snack or a link to a repository

https://github.com/software-mansion-labs/private-mind

### React Native Executorch version

0.9.2

### React Native version

0.83.4

### Platforms

Android

### JavaScript runtime

_No response_

### Workflow

React Native

### Architecture

_No response_

### Build type

_No response_

### Device

Real device

### Device model

Google Pixel 10 / Tensor G5 (Android 17, CP2A.260805.005)

### AI model

gemma4-e2b-multimodal, Vulkan export (tag v0.9.0)

### Performance logs

GPU power held a steady ~1.29 W (`S2S_VDD_GPU`) for all 12 minutes — working,
not deadlocked. `dumpsys gpu` during the stall:

```
Memory snapshot for GPU 0:
Global total:      4007948288
Proc 11331 total:  3602255872     <- the app
Proc 610 total:     131170304     <- surfaceflinger
```

The app holds 3.6 GB of a 4.0 GB global GPU budget — the whole model is resident
on the GPU, consistent with the display starving rather than the GPU hanging.

The allocation outlives the process: after `am force-stop`, with `pidof`
returning nothing, the snapshot still reports those 3.6 GB against the dead PID.

### Acknowledgements

Yes

kfaracik:
Closes #344. Takes `react-native-executorch` from 0.9.2 to 0.10.2 through its `./legacy` subpath, and pins the native artifacts to the `v0.10.4-libs` pre-release, which carries the PowerVR prepack cherry-pick behind software-mansion/react-native-executorch#1475.

## The cascade from #344, resolved

| blocker | resolution |
| --- | --- |
| reanimated 4.3.1 pins worklets 0.8.x, RNE 0.10 wants `>=0.10.0 <0.13.0` | reanimated → 4.5.5 with worklets 0.10.4 — the smallest step that reaches the range and still declares RN 0.83 |
| `@react-native-rag/executorch` peer-requires RNE ^0.9.0 and has no 0.10 release | dropped; `LFMEmbeddings` implements react-native-rag's `Embeddings` directly over `TextEmbeddingsModule` (the adapter was forty lines we already extended) |
| every import moved under `./legacy` | taken as-is, so the diff is import specifiers; the `use<Task>` / `create<Task>` rewrite stays a separate decision |

Two knock-on fixes came out of it: `react-native-executorch-expo-resource-fetcher` also needs its `/legacy` entry point (its 0.10 entry reaches for the Expo SDK 56 file-system API and throws during module evaluation on SDK 55), and `GEMMA4_E2B` / `GEMMA4_E2B_MM` no longer exist as grouped objects, so they are rebuilt here from the per-backend URL constants, which still resolve to the same v0.9.0 files. `WHISPER_TINY_EN` is the one model whose files genuinely move, so speech-to-text downloads once more after the upgrade.

## The crash that looked like ExecuTorch and was not

On 0.10.2 the app aborted on the first streamed token:

```
F libc: Fatal signal 6 (SIGABRT) in tid (mqt_v_js)
Abort message: uncaught exception facebook::jni::JniException
#08 libworklets.so
```

It arrives right after a model answers, so it reads as an inference failure. It is the worklets Bundle Mode plugin: `workletizableModules` was the pre-0.10 spelling of what is now `importForwarding.moduleNames`. Babel ignores plugin options it does not recognise, so the rename cost nothing at build time and everything at run time — `remend` never reached the worklet runtime, and streamdown could not render. 0.11.4 crashed identically, so the worklets version was never the problem.

## The libs pin

0.10.2 pins its native artifacts to `v0.10.0-libs` via `nativeLibsVersion`, and that build predates pytorch/executorch#22147: its Vulkan backend still carries `no_op.glsl` and has no `Context::register_barrier`. `v0.10.4-libs` has the cherry-pick, plus the upstream Vulkan work in pytorch/executorch#22909 and a core fix for programs whose constant segment holds only zero-length tensors.

`nativeLibsVersion` is a field in the dependency's own package.json with no per-project override, so it moves by patch-package; the package's own postinstall has already fetched `v0.10.0-libs` by the time patches apply, so `postinstall` runs the downloader once more afterwards. It always extracts, and its cache directory is keyed on the version it reads, so the second pass replaces `third-party/` with the pinned artifacts.

Verified from an empty `node_modules`, and again inside the built APK:

```
lib/arm64-v8a/libvulkan_executorch_backend.so
  sha256  dec05d05aca56fb71eaf8c87b264cfb04dee7a191bfa0de336b92587c9e03d7e   (= the release asset)
  strings | grep -c no_op             0
  nm -DC | grep -c register_barrier   1
```

Both readings are inverted from what the issue measured on `v0.10.0-libs` (18 and 0).

## What the constants actually do across the upgrade

Resolving every model constant the app imports, from both packages, on both platforms:

| | |
| --- | --- |
| 13 of the 14 model constants | byte-identical between 0.9.2 and 0.10.2 |
| `WHISPER_TINY_EN` | files move `v0.9.0` → `v0.10.0`, so speech-to-text downloads once more |
| `GEMMA4_E2B`, rebuilt here | identical to the 0.9.2 object, field for field, on Android and iOS |
| `GEMMA4_E2B_MM`, rebuilt here | same three sources; 0.9.2 also carried `capabilities` and `audioConfig`, which the app never read — `llmStore` builds its own `'custom'` descriptor from the three paths and passes `capabilities` from `model.vision`, exactly as on `main` |

## Testing

Release build, Galaxy S20 FE (SM-G781B, Android 13), installed in place over 1.3.0:

- cold start, no abort — the `libworklets.so` SIGABRT above is gone
- LFM 2.5 VL 450M and LFM 2.5 1.2B downloaded through `ExpoResourceFetcher/legacy`
- text reply streams and renders: *"The capital of France is Paris."*
- image reply through the multimodal legacy path: a red circle → *"The shape in this image is a circle, and its color is red."*, in seconds
- document RAG end to end on the rewritten `LFMEmbeddings`: 415 MB embedding model downloaded, a Polish test document indexed, and the answer quotes it correctly with a Sources chip
- web search end to end on 1.2B: planner → fetch → grounded answer with a source badge, 38 s
- switching model variants from the chat header — the path behind the #291 SIGABRT — does not crash on reanimated 4.5.5
- keyboard open/close and message-list scrolling behave: the bar tracks the keyboard, the scroll-to-latest button appears and no layout gap opens
- speech input loads: Whisper re-downloads at the new v0.10.0 URLs and the live-transcript recorder starts (no mic input to assert a transcription)
- `logcat -b crash` and the `killinfo` / `am_proc_died` event records carry nothing for the app across the whole session

`yarn typecheck`, `yarn lint` (0 errors) and 144 suites / 2451 tests pass.

Earlier in the branch, on a Pixel 10: model download, a text reply, and an image answered correctly by LFM 2.5 VL — see 9113610, 201d204 and 955512f.

## What this does not answer yet

**Whether the pinned libs fix software-mansion/react-native-executorch#1475 is still open.** Gemma 4 is `Incompatible` on the S20 FE (2.50 GB / 4.00 GB against 5.5 GB of RAM), so neither Gemma variant can be downloaded there and the Vulkan multimodal path — the one that starved the display for twelve minutes — was never exercised in this round. That run needs the Pixel 10.

Transcription itself is unasserted, and not re-run here: an iOS build (the lockfile was regenerated in 8eace8e, and `v0.10.4-libs` also replaces the iOS core and MLX frameworks — the MLX change bounds the lazy graph and cuts peak memory), and speech-to-text, which will re-download Whisper on first use.

cc @msluszniak — the libs are yours, from `@ms/separate-backends-1.4.1`.


msluszniak:
Before I'll proceed with review, you need to verify yourself that the new libs fix the problem with Pixel. I don't have any Pixel so I'm not able to do that.

kfaracik:
Closing. The upgrade is not worth taking right now, and the branch stays on the
remote for whoever picks this up next.

## Why

The payoff is gone. This bump existed to fix the Pixel 10 Vulkan stall from
software-mansion/react-native-executorch#1475 by pinning `v0.10.4-libs`.
#339 fixes that for our users without touching the library: it moves Gemma 4 VL
off Vulkan on Android through the remote catalog, so it reaches every 1.3.0
install without a store release. After #339 no multimodal model of ours enters
the Vulkan path at all, which is the path the cherry-pick repairs.

The cost is total on iOS and now precisely measured. On 0.10.2 both Gemma 4
variants fail on the first send:

```
name=Error  code=1  msg=Failed to generate text
```

`code=1` is `executorch::runtime::Error::Internal`, thrown at
`legacy/cpp/rnexecutorch/models/llm/LLM.cpp:85` when `runner_->generate()`
returns non-Ok. The model loads — a load failure raises a different message.

A/B on one iPhone 17 (iOS 26.6.2), same model files on disk, same app code,
same prompt:

| build | Gemma 4 - 2B |
| --- | --- |
| 0.9.2 (`main`) | full, correct answer |
| 0.10.2 (this branch) | `code=1` instantly |

Qwen 3 - 1.7B generates normally on the same 0.10.2 build, so the boundary is
the backend, not the library as a whole: RNE's registry maps `gemma4_e2b` to
`ios: 'mlx'`, while Qwen, LLaMA, LFM and Bielik resolve to XNNPACK. Gemma is the
only family we ship on iOS that reaches MLX, and Gemma 4 VL fails identically.

Ruled out on our side: the MLX backend is linked (5619 MLX symbols in the
binary; `enableMlx` defaults on in the podspec), the descriptor we pass matches
the registry's `mlx` variant field for field, and `modelName: 'custom'` is the
sanctioned route — `LLMModule.fromCustomModel` passes the same value and the
docs say it is for telemetry only.

Worth raising with the library separately: the failure prints nothing. The
device console carries `ET_LOG` at INFO (tokenizer regex compilation, the PCRE2
fallback), but `generate()` returns `Internal` silently, with no `E` or `W` line
before it.

## Verified on Android, and worth keeping

A release APK from this branch on the Pixel 10 from #1475 (Android 17,
`CP2A.260805.005`): a 768x768 image answered correctly in ~20 s, `screencap`
during prefill returned in 1.23 / 0.44 / 0.37 s, and 30k logcat lines hold no
`waitForever`, `pvrsrvkm`, `GUILTY_OVERRUNING`, GPU reset, ANR or lmkd kill.
The pinned Vulkan backend in the APK checks out: `libvulkan_executorch_backend.so`
sha256 `dec05d05aca56fb71eaf8c87b264cfb04dee7a191bfa0de336b92587c9e03d7e`,
`no_op` 0, `register_barrier` 1 — both inverted from `v0.10.0-libs`.

That answers the question asked in review, independently of this PR landing.

## What unblocks a second attempt

A react-native-executorch release where MLX generation works on iOS. Failing
that, the same lever #339 used on Android works here: Gemma has an XNNPACK
export (`e2b/xnnpack/gemma_4_e2b_xnnpack_8da4w.pte`, 2.64 GB at `v0.9.0`) and
the registry knows the `xnnpack` variant, so pointing the iOS `modelPath` at it
in the catalog avoids MLX entirely, at the cost of CPU-speed generation and a
re-download.

The branch keeps work that has nothing to do with the blocker and should not be
rebuilt from scratch: the `workletizableModules` to `importForwarding.moduleNames`
rename, reanimated 4.5.5 with worklets 0.10.4, `LFMEmbeddings` implemented
directly over `TextEmbeddingsModule` instead of the dead
`@react-native-rag/executorch` adapter, the `/legacy` entry point for the
resource fetcher on Expo SDK 55, and the `nativeLibsVersion` pin through
patch-package.

It does not conflict with #339: the two overlap only in
`constants/default-models.ts` and `utils/promptUtils.ts`, `git merge-tree`
reports no conflict markers, and #339's Android override sits at the call site,
so it survives this branch rebuilding the `GEMMA4_E2B_MM` object.

